PR #9, opened 2015-03-21
Recommend virtualenv for pip

kuba:
As per request in letsencrypt/acme-spec#79, I'm sending the patch upstream :) In short: just `pip install` will fail (permissions error), and `sudo pip` pollutes system-wide packages.

FYI there is also `pip install --user` that installs to `~/.local`, but in my opinion, in practice, `virtualenv` works a lot better. I leave it up to you, to evaluate which one is better for this project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@
 /*-[0-9][0-9].xml
 .refcache
 .i-d-template
-
+venv/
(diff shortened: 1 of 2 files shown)


martinthomson:
Thanks for doing that.  I think that perhaps I will enumerate the options, as you have in your intro here.

PR #18, opened 2015-10-20
Fix sed command to extract Draft Title from XML

madofo:
This was failing for me on Mac OS X and I tracked it down to these missing semicolons.

At http://sed.sourceforge.net/sedfaq.html I found this: "Following an address or address range, sed accepts curly braces '{...}' so several commands may be applied to that line or to the lines matched by the address range. On the command line, semicolons ';' separate each instruction and must precede the closing brace."


martinthomson:
Thanks, that can't have been easy to track down.

PR #20, opened 2015-11-26
updated with kramdown-rfc2629 error

nrooney:


martinthomson:
I think that [this](https://github.com/martinthomson/i-d-template/blob/master/doc/SETUP.md#kramdown-rfc2629) is clear enough.

PR #31, opened 2016-03-04
This is actually required at least to make JSEP work

ekr:


martinthomson:
I am surprised at the `git add` change.  The `.gitignore` in that branch should be empty or non-existent.

PR #32, opened 2016-03-16
Improve CygWin instructions

MikeBishop:
I just clean-installed the computer and had to rebuild my CygWin environment to generate the last couple days of drafts.  I'd like to make sure the info I needed to get back to a workable state is in the docs, so I don't have to re-research the next time I flatten the machine.

Oh, and it might help someone else, too.  ;-)

### DIFF
--- a/doc/SETUP.md
+++ b/doc/SETUP.md
@@ -18,7 +18,16 @@ Windows users will need to use [Cygwin](http://cygwin.org/) to get `make`.
 
 All systems require [xml2rfc](http://xml2rfc.ietf.org/).  This
 requires [Python](https://www.python.org/).  The easiest way to get
-`xml2rfc` is with `pip`.
+`xml2rfc` is with `pip`, which is part of the `python-pip` package
+on most distributions.
+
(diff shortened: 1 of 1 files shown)


doc/SETUP.md
```diff
@@ -18,7 +18,16 @@ Windows users will need to use [Cygwin](http://cygwin.org/) to get `make`.
...
 All systems require [xml2rfc](http://xml2rfc.ietf.org/).  This
 requires [Python](https://www.python.org/).  The easiest way to get
-`xml2rfc` is with `pip`.
+`xml2rfc` is with `pip`, which is part of the `python-pip` package
+on most distributions.
+
+On Cygwin, you'll need to install `pip` directly:
+
+```sh
+$ lynx -source https://bootstrap.pypa.io/get-pip.py | python
```
martinthomson: lynx?  isn't curl better?

martinthomson:
Glad to take the change.  I think that `lynx -source` can be replaced with simply `curl`.

PR #34, opened 2016-03-21
Make diffs happen in the right order

ekr:
Presumably you want the forward diff and not the backward diff


martinthomson:
Order matters apparently.

PR #36, opened 2016-03-25
Fix PR detection

ekr:
MT, TRAVIS_PULL_REQUEST is a number so the later test against true is wrong

### DIFF
--- a/id.mk
+++ b/id.mk
@@ -34,7 +34,12 @@ CI_REPO = $(word 2,$(subst /, ,$(TRAVIS_REPO_SLUG)))$(CIRCLE_PROJECT_REPONAME)
 ifeq (true,$(CI))
 CI_REPO_FULL = $(CI_USER)/$(CI_REPO)
 endif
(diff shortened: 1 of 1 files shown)


id.mk
```diff
@@ -34,7 +34,12 @@ CI_REPO = $(word 2,$(subst /, ,$(TRAVIS_REPO_SLUG)))$(CIRCLE_PROJECT_REPONAME)
 ifeq (true,$(CI))
 CI_REPO_FULL = $(CI_USER)/$(CI_REPO)
 endif
-CI_IS_PR = $(if $(CI_PULL_REQUESTS),true,$(if $(TRAVIS_PULL_REQUEST),$(TRAVIS_PULL_REQUEST),false))
+ifeq (false,$(TRAVIS_PULL_REQUEST))
```
martinthomson: If it's a number, then test will fail on travis and everywhere else.  You only needed to change the line above to:

```
CI_IS_PR = $(if $(CI_PULL_REQUESTS),true,$(if $(TRAVIS_PULL_REQUEST),true,false))
```

PR #42, opened 2016-06-08
partial support rfc2629.xslt

reschke:


main.mk
```diff
@@ -41,6 +41,14 @@ endif
 %.txt: %.xml
 	$(xml2rfc) $< -o $@ --text
 
+%.html: %.jrxml rfc2629xslt/rfc2629.xslt
+    $(xsltproc) rfc2629xslt/rfc2629.xslt $< > $@
+
+%.txt: %.jrxml rfc2629xslt/clean-for-DTD.xslt
+    $(xsltproc) rfc2629xslt/clean-for-DTD.xslt $< > $@.cleaned.xml
+    $(xml2rfc) $@.cleaned.xml -o $@ --text
+    rm $@.cleaned.xml
```
martinthomson: `-rm -f $@.cleaned.xml`

main.mk
```diff
@@ -41,6 +41,14 @@ endif
 %.txt: %.xml
 	$(xml2rfc) $< -o $@ --text
 
+%.html: %.jrxml rfc2629xslt/rfc2629.xslt
+    $(xsltproc) rfc2629xslt/rfc2629.xslt $< > $@
+
+%.txt: %.jrxml rfc2629xslt/clean-for-DTD.xslt
```
martinthomson: If you formulate this as two separate rules (one for translating .jrxml to .cleaned.xml; one for translating to text, then you don't need to include the rm step because make will do it for you.

martinthomson:
If you want to use a different file extension, then you need to modify id.mk to include jrxml in the wildcard lookup.  This also needs changes to support CI.

main.mk
```diff
@@ -41,6 +41,14 @@ endif
 %.txt: %.xml
 	$(xml2rfc) $< -o $@ --text
 
+%.html: %.jrxml rfc2629xslt/rfc2629.xslt
```
martinthomson: Where is the rule for creating `rfc2629xslt/rfc2629.xslt` ?

reschke: It's currently in the top-level make file (the one that includes this) -- I guess I should move it down?

main.mk
reschke: ok

main.mk
reschke: ok

reschke:
I don't *wish* to use a different file extension; I just didn't know another way how to specify which generation path should be taken (we actually could peek into the file and check for the presence of extensions...).

martinthomson:
Move the includes into config.mk.  If it isn't in this repo, people won't get it and the top-level makefile doesn't need those rules.

> I don't wish to use a different file extension

If you don't want to peek, we could trial use of this guarded by a variable (and set that variable based on peeking later).  The rules regarding conversion to text and html can be put in a conditional block.

```
ifeq (1,$(JULIANS_GOOD_STUFF))
# your new rules, but with s/jrxml/xml/
else
# any existing rules you want to replace
endif
```

I'm willing to try this out myself because I know that your stuff is better in a few ways.  Later, we might progressively enable this: maybe conditional on xsltproc being available.

BTW, any include rules should be behind this variable as well.

martinthomson:
@reschke should I be looking at this or #43?

reschke:
Please use #43

PR #51, opened 2016-11-24
Adding Julian Reschke's xslt as an option

martinthomson:
@reschke, can you test that this works for you?

Test this by putting `USE_XSLT:=1` in your Makefile.  If you have CI setup, then you also have to modify circle.yml/.travis.yml to include the install for xsltproc (unless you don't need that on those machines, I haven't checked).

TODO: Modify setup.mk to accept USE_XSLT=1 and then automatically modify those files as needed.

Note: still reluctant to make this a default since it requires more software; it will definitely be slower.

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -51,8 +51,28 @@ endif
 %.txt: %.xml
 	$(xml2rfc) $< -o $@ --text
 
+ifeq (1,$(USE_XSLT))
+XSLTDIR ?= $(LIBDIR)/rfc2629xslt
+$(LIBDIR)/rfc2629.xslt:	$(XSLTDIR)/rfc2629.xslt
+	$(xsltproc) $(XSLTDIR)/to-1.0-xslt.xslt $< > $@
+
+$(LIBDIR)/clean-for-DTD.xslt: $(LIBDIR)/rfc2629xslt/clean-for-DTD.xslt
+	$(xsltproc) $(XSLTDIR)/to-1.0-xslt.xslt $< > $@
+
+$(XSLTDIR)/clean-for-DTD.xslt $(XSLTDIR)/rfc2629.xslt: $(XSLTDIR)
+$(XSLTDIR):
+	git clone --depth 10 -b master https://github.com/reschke/xml2rfc $@
+
+%.cleanxml: %.xml $(LIBDIR)/clean-for-DTD.xslt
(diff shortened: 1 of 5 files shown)


reschke:
FWIW, xsltproc/rfc2629.xslt seems to be significantly faster than xml2rfc (tests with quic-tls.md):

xml2rfc.tcl 1.3s
xml2rfc.py 1.1s
xsltproc rfc2629.xslt 0.2s
saxon rfc2620.xslt 2.8s

(note Saxon includes Java startup)

main.mk
```diff
@@ -51,8 +51,28 @@ endif
...
+	$(xsltproc) $(XSLTDIR)/to-1.0-xslt.xslt $< > $@
+
+$(LIBDIR)/clean-for-DTD.xslt: $(LIBDIR)/rfc2629xslt/clean-for-DTD.xslt
+	$(xsltproc) $(XSLTDIR)/to-1.0-xslt.xslt $< > $@
+
+$(XSLTDIR)/clean-for-DTD.xslt $(XSLTDIR)/rfc2629.xslt: $(XSLTDIR)
+$(XSLTDIR):
+	git clone --depth 10 -b master https://github.com/reschke/xml2rfc $@
+
+%.cleanxml: %.xml $(LIBDIR)/clean-for-DTD.xslt
```
reschke: this needs a dependency on rfc2629.xslt as well

style.css
```diff
@@ -198,6 +202,9 @@ ol, ul, li, p {
 li, p {
   margin-left: 0;
 }
+address {
+  font-style: normal;
+}
 
 .github-fork-ribbon-wrapper {
   display: none;
```
martinthomson: Yeah, xml2rfc still produces braindead HTML for the toc.  And you use `<address>`, which causes address to be italicized in many browsers, against all good taste.

martinthomson:
As for speed, as an editor, I would be amenable to a PR that enabled xsltproc on the quic repo :)

style.css
reschke: As per <https://greenbytes.de/tech/webdav/draft-iab-html-rfc-03.html#element.address>, approved by the IAB. That's why I set override it in the CSS.

reschke:
Isn't the plan to let the quic repo use *this*?

PR #56, opened 2016-12-01
Don't try to build diff if no previous version

MikeBishop:
I think this should work without breaking anything else.  Preparatory to adding draft_diffs to the ghpages recipe.


MikeBishop:
Cancel that -- just realized it breaks further down.  Need to be a little savvier....

PR #57, opened 2016-12-01
Don't build diffs if no previous version

MikeBishop:
Let's try this again.  I think this works now.


martinthomson:
Wow, it works, but it makes the code even harder to understand.

MikeBishop:
Do you really think that's possible at this point?  I spent about 45 minutes just *understanding* it enough to make a 2-3 line change!

martinthomson:
As I said, it's the black arts in there.  I think that I prefer the new PR I put up for simplicity.  At least it's only shell scripting that way.

martinthomson:
BTW, 45 min well spent.

Issue #58, opened 2016-12-01
"make diff" shows the wrong draft numbers


MikeBishop:
The "-latest" to "-nn" transformation is applied to the current version, not the previous version.  This results in -00 versus -latest comparisons appearing to be -latest versus -01.

PR #60, opened 2016-12-01
Use .target.mk for diff

martinthomson:
@MikeBishop does this work for #58 ?


MikeBishop:
It omits the fix in mine from the second commit (ca778a8) to use -latest instead of the unsubmitted-future-draft number, but otherwise achieves the same thing.

PR #61, opened 2016-12-01
Diff version numbers

MikeBishop:
I think this achieves the same fix with more simplicity.  No, the code's not any *easier* to read, but I don't think it's *harder* per se....


MikeBishop:
Obsolete, since you're now using the shell script instead.

PR #64, opened 2016-12-02
rfcdiff takes &, not ^

MikeBishop:
Found this while incorporating it on QUIC's page


martinthomson:
I just assumed that since Jim's last patch was for that specific character, he knew something I didn't.

jimsch:
No Jim is an idiot who can't type

PR #65, opened 2016-12-07
Eol breaks sed

MikeBishop:
Think I’ve figured out the random sed breaking issue.  If I open addstyle.sed and force all line endings to be LF, drafts build.  If I force them to be CRLF, sed breaks complaining about extra characters.  But git is set to be EOL-agnostic, so changes to line endings don’t show up as changes.  Bah.

This adds a .gitattributes file that says .sed files MUST always use LF line endings, even on Windows.


martinthomson:
Awesome, that helps a lot.

PR #72, opened 2017-01-13
Bring your whitespace change back to template

MikeBishop:
...and squash the complaints that the updated Makefile is out-of-date.  :-)


martinthomson:
I think that I'd rather just suppress the warning: it should be allowed to add lines to the makefile.  See 04b7d9c.

PR #73, opened 2017-01-19
Use git hooks to prevent committing anything that won't make

MikeBishop:
Building on your Makefile-reject of whitespace in QUIC, this uses git client hooks to check that a commit *will* make successfully before allowing you to commit.  (You can always bypass with `git commit --no-verify`, of course, and it tells you that if it fails.)

It took a few twists before I got it working for both setup and existing repos, so feel free to squash it together.  I think this should be in good shape now, but please bang on it a little.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
+
+exec 1>&2
+
 # No tests to update gh-pages
-BRANCH=$(git symbolic-ref HEAD 2>/dev/null)
-[ $BRANCH = "refs/heads/gh-pages" ] && exit 0
+BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null)
+[ "$BRANCH" = "gh-pages" ] && exit 0
 
 git stash save -k -q
 make
@@ -14,5 +17,3 @@ then
   echo "To commit anyway, run \"git commit --no-verify\""
   exit 1
 fi
-
-exit 0
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+# No tests to update gh-pages
+BRANCH=$(git symbolic-ref HEAD 2>/dev/null)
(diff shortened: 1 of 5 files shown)


update.mk
```diff
@@ -30,5 +30,6 @@ update:
 	  [ -z "$(comm -13 $$i $(LIBDIR)/template/$$i)" ] || \
 	    echo $$i is out of date, check against $(LIBDIR)/template/$$i for changes.; \
 	done
+	@[ -L .git/hooks/pre-commit ] || ln -s ../../lib/pre-commit.sh .git/hooks/pre-commit; \
```
martinthomson: You don't need the trailing ;\

pre-commit.sh
```diff
@@ -0,0 +1,18 @@
...
+[ $BRANCH = "refs/heads/gh-pages" ] && exit 0
+
+git stash save -k -q
+make
+RESULT=$?
+git stash pop -q
+
+if [ $RESULT -ne 0 ]
+then
+  echo "Commit refused -- documents don't build successfully."
```
martinthomson: These need to go to stderr. `exec 1>&2` at the top might do the trick.

pre-commit.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# No tests to update gh-pages
+BRANCH=$(git symbolic-ref HEAD 2>/dev/null)
+[ $BRANCH = "refs/heads/gh-pages" ] && exit 0
+
+git stash save -k -q
+make
```
martinthomson: I think that it's only necessary to build one version, though it probably won't save much time overall.  Maybe a `make check` target would save some time.

MikeBishop: If xml2rfc had a `--validate-only` option, that would be ideal.  Let kramdown generate the XML, make sure it's valid with xml2rfc, and stop there with no files output.  Failing that, I'd guess that text-only is probably a little cheaper than HTML as well, but I don't know by how much.  (Tonight, I'm getting timeouts on all attempts to refresh references, so that dominates everything.)

PR #79, opened 2017-01-31
Prune expired artifacts from gh-pages

### LINKED ISSUE #68: gh-pages pollution
Check out the gh-pages branch of the QUIC repo.  The old versions of the documents pre-rename are there, as well as leftovers from every defunct branch that ever built gh-pages.  Should there be some pruning to make sure we drop old build artifacts, or does it not matter since those things are going to live forever in git anyway, whether they're in the current state or not?

MikeBishop:
Fixes #68.  Two key cleanup actions:

- When building any branch, if a draft no longer exists in that branch, remove it from that branch's directory.  This makes renames take effect, rather than leaving [old copies of documents](https://quicwg.github.io/base-drafts/draft-ietf-quic-http-mapping.html) that might confuse someone.
- When building master, do an extra check for [deleted branches](https://quicwg.github.io/base-drafts/http-name/).  Remove them when:
    - They no longer exist on any remote (which will only be origin for CI)
    - The documents were last built more than 30 days ago

I would have made the second one "the branch was deleted more than 30 days ago," but it appears that once a branch is deleted, git retains no record that it ever existed.  Please pay particular attention to line 109; I think I read somewhere that `date` has different parameters on Mac, so this line may also need to be variable based on the system, but I don't know what that format is.

### DIFF
--- a/ghpages.mk
+++ b/ghpages.mk
@@ -71,17 +71,6 @@ $(GHPAGES_TMP): fetch-ghpages
 
 PUBLISHED := index.html $(drafts_html) $(drafts_txt)
 
-ifeq (master,$(SOURCE_BRANCH))
-ifeq (true,$(PUSH_GHPAGES))
-EXTANT_BRANCHES := $(subst refs/heads/,,$(foreach remote,$(shell git remote),$(shell git ls-remote --heads $(remote) | cut -f 2)))
-OBSOLETE_DIRECTORIES := $(shell git ls-tree -t --name-only gh-pages)
-OBSOLETE_DIRECTORIES := $(filter-out circle.yml .gitignore,$(OBSOLETE_DIRECTORIES))
-OBSOLETE_DIRECTORIES := $(filter-out $(EXTANT_BRANCHES),$(OBSOLETE_DIRECTORIES))
-OBSOLETE_DIRECTORIES := $(filter-out $(PUBLISHED),$(OBSOLETE_DIRECTORIES))
-endif
-endif
-OBSOLETE_DIRECTORIES ?=
-
 .PHONY: ghpages gh-pages
 gh-pages: ghpages
 ghpages: $(PUBLISHED)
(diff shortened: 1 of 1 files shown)


ghpages.mk
```diff
@@ -84,6 +99,22 @@ ifeq (true,$(PUSH_GHPAGES))
 ifneq (,$(TARGET_DIR))
 	mkdir -p $(GHPAGES_TMP)/$(TARGET_DIR)
 endif
+
+	@EXCESS_FILES="$(addprefix $(TARGET_DIR),$(filter-out $^,$(notdir $(foreach ext,*.html *.txt,$(wildcard $(GHPAGES_TMP)/$(TARGET_DIR)/$(ext))))))"; \
+	if [ -n "$$EXCESS_FILES" ]; \
+		then git -C $(GHPAGES_TMP) rm -f --ignore-unmatch -- $$EXCESS_FILES; \
+	fi
```
martinthomson: I wonder if this can't be simplified greatly by making this just:
```
git -C $(GHPAGES_TMP) rm -fq --ignore-unmatch -- $(GHPAGES_TMP)/$(TARGET_DIR)/*.txt $(GHPAGES_TMP)/$(TARGET_DIR)/*.html
```
The files that are being added here will be added back.  I just checked and `git rm` followed by `git add` works fine.

ghpages.mk
```diff
@@ -84,6 +99,22 @@ ifeq (true,$(PUSH_GHPAGES))
...
+
+ifneq (0,$(words $(OBSOLETE_DIRECTORIES)))
```
martinthomson: This is nice from an execution performance standpoint, but I think that it would be better to avoid the work of constructing the OBSOLETE_DIRECTORIES variable unless you are running this target.  If you take the above into account, this gets easier.  In shell (which needs more quoting):

```make
@cutoff=$$(date +%s -d '-30 days'); \
comm -13 <(git branch -a | sed -e 's,.*[ /],,' | sort | uniq) <(ls $(GHPAGES_TMP) | sed -e 's,.*/,,')) | while read obsolete; do \
  [ ! -d "$$obsolete" ] || \
    [ "$$cutoff" -le "$$(git -C $(GHPAGES_TMP) log -n 1 --format=%ct "$$obsolete")" ] || \
    echo "Remove obsolete '$$obsolete'" && git rm -rfq -- "$$obsolete"; \
done
```

You can then run the cleanup always.  The only trick would be to move the command that makes the target directory after this cleanup in case the directory is removed in this cleanup.  That is, the order is:
1. obsolete directory cleanup
2. mkdir $(TARGET_DIR)
3. cleanup $(TARGET_DIR)
4. add updated files
5. commit
6. push

ghpages.mk
```diff
@@ -68,9 +71,21 @@ $(GHPAGES_TMP): fetch-ghpages
 
 PUBLISHED := index.html $(drafts_html) $(drafts_txt)
 
+ifeq (master,$(SOURCE_BRANCH))
+ifeq (true,$(PUSH_GHPAGES))
+EXTANT_BRANCHES := $(subst refs/heads/,,$(foreach remote,$(shell git remote),$(shell git ls-remote --heads $(remote) | cut -f 2)))
```
MikeBishop: The difference, I think, is that ls-remote pulls live data from the remotes, while yours is only current as of the last fetch.  Also, remote-tracking branches aren't cleaned up automatically.

Though I could fetch all remotes and prune all remotes first, I was trying to avoid side-effects of gathering the data.

martinthomson: Ugh, that's a double-reason not to use this: it will block a regular `make` on the network.

I have no problem with having this not use remotes that weren't recently fetched.  We already check if the directory is old, so the only risk is that the remote has a big gap.  On CI, we don't have any other remotes, so it's going to prune aggressively there anyway.

MikeBishop:
I hadn't realized all this would run for other `make`s; that's a deal-breaker, I think.  I've taken your suggestion and moved a lot of this to shell script logic instead.  See what you think now?

PR #81, opened 2017-02-01
pass the correct source file to rfc2629.xslt

reschke:


martinthomson:
Thanks.  I'll have to think some more about .cleanxml and its use during draft submission.

PR #82, opened 2017-02-02
Cache env variable isn't defined when this section is parsed

MikeBishop:
Your last commit is generating a warning on Circle that the requested cache directory doesn't exist.  I *think* this fixes it.


MikeBishop:
Still no.  The directory doesn't exist when the cache is snapped -- I think this is just broken entirely.

martinthomson:
Oh well, I'll just delete the caching rule.

PR #83, opened 2017-02-04
Set name/email at repo setup

MikeBishop:
Rather than remembering to set this on each git command where it matters, why not just set them in the repo config when the repo is created?


MikeBishop:
Hm.  This seems to break in the gh-pages flow; that may not currently work properly if the user's config isn't global.  I'll see about fixing that, and then come back to this change.

martinthomson:
So the point of the `-c user.name=Foo` on the commit was to avoid this.  Is this failing in your version of git?  It's working well enough for me.

MikeBishop:
Incorporated into #84.

PR #85, opened 2017-02-08
Only stash (and apply) if there are uncommitted changes

MikeBishop:
I noticed some odd artifacts of an attempted stash application after committing, and realized that there's a bug in the following situation:

- You commit with nothing in your working directory (so `git stash` doesn't create a new stash)
- You have old stashes that are still hanging around (so `git stash pop` *does* apply a stash)

This checks whether there are unstaged changes and only runs the stash/apply if there were.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -11,7 +11,7 @@ if [ "$BRANCH" = "gh-pages" ] ||
 fi
 
 STASHED=0
-if test `git status --porcelain | grep '^[A-Z]' | wc -l` -gt 0; then
+if [ `git status --porcelain | grep '^[A-Z]' | wc -l` -gt 0 ]; then
   git stash save -k -q
   STASHED=1
 fi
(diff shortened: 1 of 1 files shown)


pre-commit.sh
```diff
@@ -5,14 +5,23 @@ exec 1>&2
...
 BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null)
 if [ "$BRANCH" = "gh-pages" ] ||
+   [ "$BRANCH" = "gh-issues" ] ||
    [ -e .git/MERGE_HEAD ]; then
      exit 0
 fi
 
-git stash save -k -q
+STASHED=0
+if test `git status --porcelain | grep '^[A-Z]' | wc -l` -gt 0; then
```
martinthomson: A mishmash of [ and test is a bit strange.

martinthomson: Here's a different approach that might work:

list changed drafts
copy affected drafts to temporary names
make "${affected_drafts}.txt"
rm copies

MikeBishop: No, because you can have staged a valid version, and since made a breaking change that isn't part of the commit.  I don't think that works if you just ignore files with changes -- what you want is to check that whatever's currently staged for commit will build successfully.

I replaced the `test` with `[]`; hopefully that addresses the style concern.

PR #86, opened 2017-02-08
Filter pre-commit checks based on what is staged

martinthomson:
@MikeBishop, do you think that this is this a better version of #85?  Note that it only makes the .txt version as a secondary benefit.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -4,18 +4,25 @@ exec 1>&2
 
 # No tests to update gh-pages
 BRANCH=$(git symbolic-ref --short HEAD 2>/dev/null)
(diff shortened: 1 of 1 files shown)


MikeBishop:
There is some advantage in not building the documents that haven't been touched at all; otherwise, seems equivalent to me.

Mine leverages git's intelligence, yours leverages make's.  Both are "simple" tools that are nearly Turing-complete programming languages.  ;-)  And building only the txt files is easy enough in either -- just run `make txt` instead of `make`.

martinthomson:
There's another benefit to this: it doesn't clobber any output generated from non-staged files.  That's a minor benefit though.

Yeah, Turing complete they might be, but sane they are not.

PR #88, opened 2017-02-09
Shallow clones break gh-pages pruning

MikeBishop:
Figured out why Circle wasn't cleaning up old branches still, even though running locally does.  The template circle.yml file does a shallow fetch, so Circle never has a full history of gh-pages.  That causes all files to appear to have been created a day or so ago by the oldest commit in its clone.

This .yml change, which I've added to QUIC already, will unshallow the repo if it's shallow, so that history is available.  This shouldn't decrease perf noticeably, as Circle caches the repo between builds and won't need to re-fetch the whole history every time.

### DIFF
--- a/template/circle.yml
+++ b/template/circle.yml
@@ -6,7 +6,8 @@ machine:
 
 checkout:
   post:
-    - git fetch origin gh-pages --depth 10
+    - if [ -e .git/shallow ]; then git fetch origin --unshallow; fi
+    - git fetch origin gh-pages
 
 dependencies:
   pre:


MikeBishop:
Alternatively, some versions of git appear to support a --shallow-since= parameter, which could be set to ~45 days, or anything beyond the survival cutoff.  That might also be a reasonable choice, since it appears to be supported on Circle's machines.

MikeBishop:
...and it would appear to be *so* new that GitHub doesn't support it yet, even though Circle's client does.  We'll hold off on that improvement, then.

template/circle.yml
```diff
@@ -6,7 +6,8 @@ machine:
 
 checkout:
   post:
-    - git fetch origin gh-pages --depth 10
+    - if [ -e .git/shallow ]; then git fetch origin --unshallow; fi
+    - git fetch origin gh-pages
```
martinthomson: Do we really need the unshallow fetch?  as long as the gh-pages fetch is complete, then this should be fine, right?  I would just drop this entire step, since the makefile fetches gh-pages as needed now.

MikeBishop: The problem is, CI was running that code already and was still working with a shallow clone.  The fetch in ghpages.mk doesn't automatically fill in your history, it just gets new stuff.

MikeBishop: Oh, and while we could just move the --unshallow fetch there, that's after the cache has snapped a copy of the .git directory, so we'd be unshallowing the repo every run, rather than only once per repo this way.

martinthomson:
Ahh, caching is a good reason to do this.

PR #89, opened 2017-02-10
Circle cache

MikeBishop:
This caches gem's download directory, and declaring that you use Python actually preinstalls xml2rfc and behave, so those commands become no-ops.  (We could remove them, but it might be nice to leave the dependency explicit.)

### DIFF
--- a/template/circle.yml
+++ b/template/circle.yml
@@ -14,6 +14,9 @@ dependencies:
     - pip install xml2rfc
     - if head -1 -q *.md | grep '^\-\-\-' >/dev/null 2>&1; then gem install --no-doc kramdown-rfc2629; fi
     - if head -1 -q *.md | grep '^%%%' >/dev/null 2>&1; then go get "$mmark_src" && go build "$mmark_src"; fi
+  cache_directories:
+    - "~/.cache"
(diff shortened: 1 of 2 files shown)


template/circle.yml
```diff
@@ -14,6 +17,8 @@ dependencies:
     - pip install xml2rfc
     - if head -1 -q *.md | grep '^\-\-\-' >/dev/null 2>&1; then gem install --no-doc kramdown-rfc2629; fi
     - if head -1 -q *.md | grep '^%%%' >/dev/null 2>&1; then go get "$mmark_src" && go build "$mmark_src"; fi
+  cache_directories:
+    - "/opt/circleci/.rvm/gems"
```
martinthomson: Do you think we should set KRAMDOWN_REFCACHEDIR to ~/.cache and cache that as well?

MikeBishop:
Wouldn’t make a difference; same problem as before still applies.  The cache snaps before make runs, so it'll never be populated.

I found a Circle feature request to do an optional second snap after the test phase, but no indication that Circle is acting on it in the near future.

Sent from my Windows 10 phone

PR #90, opened 2017-03-10
Hash is noisy

MikeBishop:
When you don't have gmake, your check spews the fact to stderr when you're handling that case just fine.


martinthomson:
My bad, I usually redirect by habit.

PR #104, opened 2017-05-26
New index

martinthomson:
This builds on work by @chrysn to build a better index.  This uses a little JS to bounce people of the index page to the right place.  It also includes a rework of how the README is built during setup so that the new index can be used properly and a few other things I've been meaning to do.


chrysn:
I'm using this now in https://github.com/core-wg/resource-directory/. Thanks for integrating this and all the polishing on it.

Issues I've seen with it:
* The working group name is now capitalized in the CONTRIBUTING.md file; it needs to be lowercase for the links to work.
* Building from Travis produces very odd results [like this](https://github.com/core-wg/resource-directory/commit/d6e8e1a1d4e359d2a5ad9fa8ade6e5c21b77f01d), it seems like it fails to `cd` into somewhere and then wildcards can't be expanded and paths are all off, or something like that.

martinthomson:
I've fixed those two issues... I think.  I had no idea that travis wouldn't have realpath.  I'm running trusty on this machine and it has it in coreutils.  I added a workaround.  If you could poke your repo, it should fix the links.

chrysn:
> If you could poke your repo, it should fix the links.

Thanks, it builds fine now.

PR #118, opened 2017-12-12
Incorporate XSLT into Dockerfile

MikeBishop:
Not sure how best to test the Dockerfile change; hoping Circle will do that when I create the PR.

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -66,7 +66,7 @@ $(LIBDIR)/clean-for-DTD.xslt: $(LIBDIR)/rfc2629xslt/clean-for-DTD.xslt
 
 $(XSLTDIR)/clean-for-DTD.xslt $(XSLTDIR)/rfc2629.xslt: $(XSLTDIR)
 $(XSLTDIR):
-	git clone --depth 10 $(XSLT_CLONE_ARGS) -b master https://github.com/reschke/xml2rfc $@
+	git clone --depth 10 $(CLONE_ARGS) -b master https://github.com/reschke/xml2rfc $@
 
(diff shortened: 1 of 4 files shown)


main.mk
```diff
@@ -66,7 +66,7 @@ $(LIBDIR)/clean-for-DTD.xslt: $(LIBDIR)/rfc2629xslt/clean-for-DTD.xslt
 
 $(XSLTDIR)/clean-for-DTD.xslt $(XSLTDIR)/rfc2629.xslt: $(XSLTDIR)
 $(XSLTDIR):
-	git clone --depth 10 -b master https://github.com/reschke/xml2rfc $@
+	git clone --depth 10 $(XSLT_CLONE_ARGS) -b master https://github.com/reschke/xml2rfc $@
```
martinthomson: You need to pass this variable via the circle config

martinthomson:
Oh, you might want to change the way that this works actually.  There is a trick where you lump all the remote repos into a single bare git repository.  See https://randyfay.com/content/reference-cache-repositories-speed-clones-git-clone-reference

I think that you can use git fetch to add extra fetches to this without doing the whole remotes thing in this case.

main.mk
MikeBishop: Whoops, I thought I had.  I'll clean that up when I look at the below.

MikeBishop: I see what I did wrong.  I made my updates to the config.yml in the http-extension repo, rather than in the template.

MikeBishop:
I see what he's doing there....  If you're building a cache in one shot, as in the Dockerfile, you're correct that you could do this entirely by `git fetch` from a URL.  However, if we want to keep the property that the cache gets updated whenever there's a run (and not just when a new Docker image is built), then the remotes are easier.

PR #119, opened 2017-12-13
Support SHORTNAME_PREFIX for gh-pages

MikeBishop:
Admittedly a vanity rule to avoid breaking Mark's existing links, but this enables the creation of shortened names in gh-pages.  I can understand if some people don't want to type `draft-me-them-` all the time....


MikeBishop:
As noted on Slack, since others are looking at this PR....  .targets.mk seems to have the same drawback as doing it directly in http-extension's Makefile:  You can create new targets for the local directory very easily, but can't easily publish new things into gh-pages, which is the requirement here.  That requires knowing the temp directory make has chosen and publishing targets in that directory, which so far I've only managed to do from within ghpages.mk itself.  (Hence the eval-foreach-call setup.)

MikeBishop:
Closed via #129.

PR #120, opened 2017-12-13
fix text generation for XML inputs that use extensions

reschke:
....from rfc2629.xslt


martinthomson:
This only works when USE_XSLT=true.

martinthomson:
Fixed in f646912

PR #122, opened 2017-12-13
CSS suggestions

mnot:


style.css
```diff
@@ -31,11 +31,11 @@
 }
 
 body {
-  font: 15px "Helvetica Neue",Helvetica,Arial,sans-serif;
+  font: 14px "Helvetica Neue",Helvetica,Arial,sans-serif;
```
martinthomson: This pushes the number of characters on each line over a threshold for me.  I like the line spacing, but I think that this regresses readability.

style.css
```diff
@@ -12,12 +12,12 @@
...
     bottom: 0;
     right: 0;
-    right: calc(50vw - 525px);
+    right: calc(50vw - 550px);
     width: 300px;
     z-index: 1;
   }
   #rfc\.toc {
-    top: 15px;
+    top: 55px;
```
martinthomson: This appears to push the TOC title into the TOC itself.  It's more than 25px high.

I think that the problem here is that the xml2rfc version puts this class on the header, whereas Julian's XSLT puts it on a \<nav> container around both the title and the content.  That's annoying because it means that there will need to be a new rule to override the rules that are used for xml2rfc:

```css
#rfc\.toc { top: 55px; overflow: auto; overscroll-behavior: contain; }
#rfc\.toc ul.toc { /* override all the other rules we just added here. */ position: unset; }
ul.toc { top: 80px; overflow: auto; overscroll-behavior: contain; }
```

This is why Julian doesn't like xml2rfc.  And I totally agree with him there.

style.css
mnot: Hm. For me, 15px gives the type a colouring that's too dark (and harder to read), on OSX. I don't suppose you'd be open to making the body slightly narrower? Otherwise, I'll revert.

martinthomson: This is safari, right?  I'll take a look.

martinthomson: I think that the width is largely dictated by the monospaced elements on the page, that might mean changing the font size there to compensate.

Would `font-weight: 300;` work for you?  That causes OSX to use Helvetica Neue Light rather than Helvetica Neue, which I think looks much cleaner.  This doesn't make any apparent difference on Windows with Arial, so that's a clear win.

martinthomson:
Merged by hand, then I increased the font and added a font-weight rule.

martinthomson:
Also, I find the fact that Julian's XSLT and xml2rfc produce different HTML to be really frustrating.  The QUIC docs look really coarse now (they use h1 in the same place that Julian uses h2).

PR #127, opened 2017-12-24
Use Python 3.x

MikeBishop:
Python 3.x is the [default installed on Ubuntu going forward](https://wiki.ubuntu.com/Python), but you're installing the 2.x binaries on the Docker image.  Since I don't have 2.x and your Docker image doesn't have 3.x, there's no Python linting script I can sort out that will run on both a current default Ubuntu install and on your Circle images.

Can we bring your image up-to-date?  That seems the easiest path out.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ LABEL maintainer="Martin Thomson <martin.thomson@gmail.com>"
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates coreutils curl git make ssh libxml2-utils xsltproc \
-    python-minimal python-lxml python-pip python-setuptools python-wheel \
+    python3-minimal python3-lxml python3-pip python3-setuptools python3-wheel \
     mmark ruby \
  && rm -rf /var/lib/apt/lists/* \
  && apt-get autoremove -y && apt-get clean -y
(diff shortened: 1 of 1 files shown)


martinthomson:
Hi Mike, I had a revert this.  The problem with this is that the most recent xml2rfc release still doesn't work with python3.

jimsch:
It should work w/ python3 - what version are you using?  I have it running on my laptop w/ 3.6

martinthomson:
@jimsh: this build fails: https://circleci.com/gh/quicwg/base-drafts/3856?utm_campaign=build-failed&utm_medium=email&utm_source=notification  This would have used whatever was in pypi shortly after I merged this PR.

jimsch:
@martinthomson  python does not believe that you have a real utf-8 source file.  This is a difference between the v3 and v2 code in xml2rfc.  I cannot tell you what the invalid character is, but you do have a position in the file.  This is an input problem and not an xml2rfc problem per say.  V2 read it as a binary ascii file and did not care if everything was a legal utf-8 character or not.  I am going to guess it is an md problem, but I don't know w/o having the actual file.

jimsch:
I think I have changed my mind.  It is an xml2rfc bug and I will log it.

martinthomson:
@jimsch, I know what the problem is, it's kramdown-rfc2629 adding smart quotes.  Thanks for logging the bug.

jimsch:
Yes, but it should have been accepted by xml2rfc and processed correctly.  Currently it will not accept any input file which is not ascii even though it is supposed to.  I have a local fix that I will be sending in to deal with this.

MikeBishop:
So shall we reopen/refresh this PR and hold it until the xml2rfc fix is in?

martinthomson:
Yeah, I'll make sure the change is made once I can confirm that xml2rfc works.

PR #128, opened 2017-12-26
Modularize gh-pages

MikeBishop:
No functional change to the template (if I did it right!), but by having multiple gh-pages targets, a parent repo's Makefile can hook in and add additional functionality to gh-pages, like this:

```
ghpages: link-ghpages
commit-ghpages: | link-ghpages

link-ghpages: stage-ghpages
	@for file in $(GHPAGES_TARGET)/draft-ietf-httpbis-*.html \
					 $(GHPAGES_TARGET)/draft-ietf-httpbis-*.txt; do \
		newfilename=`echo $$file | sed -e 's/draft-ietf-httpbis-//'`; \
		ln -sf `basename $$file` $$newfilename; \
		git -C $(GHPAGES_ROOT) add `realpath --relative-to=$(GHPAGES_ROOT) $$newfilename`; \
	done;
```

Closes #119, since this is a better solution.

### DIFF
--- a/ghpages.mk
+++ b/ghpages.mk
@@ -99,10 +99,14 @@ cleanup-ghpages: $(GHPAGES_ROOT)
 	fi
 
 
-.PHONY: ghpages gh-pages
+.PHONY: ghpages gh-pages stage-ghpages commit-ghpages
 gh-pages: ghpages
-ghpages: cleanup-ghpages $(GHPAGES_ALL)
+ghpages: cleanup-ghpages commit-ghpages
+
+stage-ghpages: $(GHPAGES_ALL)
 	git -C $(GHPAGES_ROOT) add -f $(GHPAGES_ALL)
+
+commit-ghpages: stage-ghpages
 	if test `git -C $(GHPAGES_ROOT) status --porcelain | grep '^[A-Z]' | wc -l` -gt 0; then \
 	  git -C $(GHPAGES_ROOT) $(CI_AUTHOR) commit -m "Script updating gh-pages from $(shell git rev-parse --short HEAD). [ci skip]"; fi
 ifeq (true,$(PUSH_GHPAGES))


martinthomson:
Why can't you just add to `GHPAGES_PUBLISHED` (or `GHPAGES_INSTALLED` if you think that symlinks are necessary to have on the branch)?

```
GHPAGES_PUBLISHED += $(addsuffix .html,$(foreach draft,$(drafts),$(shell echo $(draft) | sed -e 's/draft-ietf-httpbis-//')))
```

And then add rules for linking `foo.html` to `draft-ietf-httpbis-foo.html`.

BTW, `realpath` isn't widely available.  You will break mac users I think.

MikeBishop:
Tried it, but I don't believe that can work.  In the current structure $(GHPAGES_INSTALLED) gets evaluated immediately as part of the rule processing with no opportunity for me to add to it before it's expanded.  (If it were internal to a recipe, I think that would work since it wouldn't actually get expanded until the recipe executed.)

If GHPAGES_PUBLISHED/GHPAGES_INSTALLED were added to by the template rather than freshly defined (assuming they're empty by default), I could throw things in there before the main Makefile imports the template, but GHPAGES_TARGET still wouldn't be (and can't be) defined until ghpages.mk gets processed, so I wouldn't know what targets to add.

This approach lets me insert a processing step in sequence and keep the necessary state in the temporary git repo rather than trying to mess with variables internal to the template.

martinthomson:
I was able to get that to work by making the following change:

```diff
-GHPAGES_PUBLISHED := $(drafts_html) $(drafts_txt)
-GHPAGES_INSTALLED := $(addprefix $(GHPAGES_TARGET)/,$(GHPAGES_PUBLISHED))
+GHPAGES_PUBLISHED += $(drafts_html) $(drafts_txt)
+GHPAGES_INSTALLED += $(addprefix $(GHPAGES_TARGET)/,$(GHPAGES_PUBLISHED))
```

And setting the extra values at the top of the Makefile.  Would that work for you?

martinthomson:
Sorry, should have said, we could define GHPAGES_EXTRA for this purpose.

MikeBishop:
If I add a GHPAGES_EXTRA into the definition of GHPAGES_INSTALLED, then define it in the Makefile, it gets swept up in the rule for GHPAGES_INSTALLED expecting it to exist in the current directory so it can be copied to ghpages.  I hadn't planned to make the links exist in the non-ghpages branch, but even if they do, if I attempt to create a definition for %.html in the current directory I'll be overwriting the template's definition and stomping on how the real drafts get created.

Argh.  This is a headache.

However, that might be sufficient if I explicitly enumerate the drafts that should be linked rather than trying to auto-link everything.

martinthomson:
Producing x.html from draft-foo-x.html is never going to be a generic rule, so I always assumed that you would be listing those rules explicitly.

PR #129, opened 2018-01-03
GHPAGES_EXTRA

MikeBishop:
Closes #128 with the simplest possible solution -- adds GHPAGES_EXTRA as suggested.  Permits anything in the working directory to be populated into gh-pages by the parent Makefile.


ghpages.mk
```diff
@@ -61,8 +61,8 @@ $(GHPAGES_TARGET): $(GHPAGES_ROOT)
 endif
 
 GHPAGES_PUBLISHED := $(drafts_html) $(drafts_txt)
```
martinthomson: Add GHPAGES_EXTRA to this line instead and save on duplication.

PR #131, opened 2018-01-10
Fix XSLT setup

### LINKED ISSUE #130: Setup fails with XSLT enabled
Following the [setup instructions](https://github.com/martinthomson/i-d-template/blob/master/doc/REPO.md) suggests just adding `USE_XSLT=true` to the setup command line, but this doesn't actually work.

> make: *** No rule to make target 'circle.yml', needed by 'setup-circle'.  Stop.

I think the addition of xsltproc to the Dockerfile actually obsoletes the setup-circle step entirely -- we should be able to just change the Makefile.

(Follow-up point -- if you *don't* use XSLT, the change to use a unified git cache is currently still keeping a cache of Julian's repo in the Dockerfile.  Does that matter?)

MikeBishop:
Fixes #130, including an initial test for XSLT.

### DIFF
--- a/setup.mk
+++ b/setup.mk
@@ -52,16 +52,10 @@ setup-files: $(TEMPLATE_FILES) README.md
 	git add $^
 
 ifeq (true,$(USE_XSLT))
-setup-master: setup-makefile setup-circle
+setup-master: setup-makefile
 .PHONY: setup-makefile
 setup-makefile: Makefile
 	sed -i~ -e '1{h;s/^.*$$/USE_XSLT := true/;p;x}' $<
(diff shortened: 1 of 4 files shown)


martinthomson:
Thanks Mike, this stuff wasn't that well tested, and now it is!

MikeBishop:
Well, at least the fact that the setup instructions work is tested.  Still no tests about actually building the draft using xsltproc.

martinthomson:
:)  We might have to rely on http-extensions to be the canary until such time as I put in the effort. :)

PR #137, opened 2018-02-08
Use shell to check contents of .targets.mk

### LINKED ISSUE #136: .targets.mk always builds
Appears to be because `$(TARGETS_DRAFTS)` and `$(TARGETS_TAGS)` are used to check whether the file needs to be rebuilt prior to the include, so the answer is always yes.  Including first, then re-including after building .targets.mk causes warning.

Maybe need to hand this off to shell to decide if there's something missing?

MikeBishop:
Fixes #136, albeit a little ugly.

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -99,7 +99,19 @@ endif
 submit:: $(drafts_next_txt) $(drafts_next_xml)
 
 .SILENT: .targets.mk
-ifneq ($(drafts) $(drafts_tags),$(TARGETS_DRAFTS) $(TARGETS_TAGS))
+
+MISSING_DRAFTS := $(shell \
+existing=`[ -r .targets.mk ] && head -1 .targets.mk`; \
+for term in $(drafts); do \
+  echo $$existing | grep -F -q $$term || (echo $$term && exit 1); \
+done;)
+MISSING_TAGS := $(shell \
+existing=`[ -r .targets.mk ] && head -2 .targets.mk | tail -1`; \
+for term in $(drafts_tags); do \
+  echo $$existing | grep -F -q $$term || (echo $$term && exit 1); \
+done;)
+
+ifneq (0,$(words $(MISSING_DRAFTS) $(MISSING_TAGS)))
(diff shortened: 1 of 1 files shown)


MikeBishop:
(Drops the time to run `make -q` on the http-extensions repo from 39 seconds to 4.9 seconds.)

martinthomson:
Thanks for taking the time to look into this.  You are right, I never managed to avoid that overhead.

On Linux, it's much faster, so I didn't notice.

I think that this is too complex though.  Take a look at my PR for this and see what you think.

MikeBishop:
As usual, we're trading off complexity versus corner cases.  Mine doesn't assume they're guaranteed to be in the same order (because I wasn't certain how deterministic that is), plus doesn't rebuild on the removal of a draft or tag, because you don't actually need to.  If order is deterministic, that might be a reasonable price to pay to have a simpler (and constant-time) check.

Otherwise, I wonder if there's a good way to meet in the middle -- could you use your approach of reading the lines into make, but then use `$(words $(filter-out $(drafts-in-file) $(tags-in-file),$(drafts) $(tags)))` and check that it's zero?

martinthomson:
Order is consistent enough.  I don't think that there are guarantees, but I could slap in a "sort" call for these variables to make it more certain.

`$(words $(filter-out ...))` is more work.  If we canonicalize, then it's probably faster. Given that `$(wildcard)` seems to produce sorted output, that seems easier to sort than it would be to do an O(N^2) operation.

martinthomson:
Oh, I already have drafts sorted.  Using `git tag --sort` might address the other concern (it seems to sort on name by default, but that's not a documented feature).

MikeBishop:
Works for me.  The trade-off of a spurious rebuild on draft deletion is probably cheap enough to pay.  I might also look and see if I can make the rebuild itself faster on WSL, but that's probably just a factor of so many drafts and tags in this particular repo.

martinthomson:
Yeah, QUIC is a pig (http-extensions is also a pig).

PR #145, opened 2018-04-03
Don't use git submodule

MikeBishop:
Doing a little exploration of what the remaining sources of slowness are in the template, it appears that each git submodule call is fairly slow (0.5-3 seconds), and there are two of them.  Replacing that with a check that .gitmodules exists && contains "path = $(LIBDIR)" appears to be a fair bit faster and is still accurate based on a quick test of a submodule repo.

The gains aren't dramatic, but the functionality appears to be equivalent.

### DIFF
--- a/id.mk
+++ b/id.mk
@@ -1,7 +1,9 @@
 ## Identify drafts, types and versions
 
-ifneq (,$(shell git submodule status $(LIBDIR) 2>/dev/null))
+ifneq (,$(wildcard .gitmodules))
+ifneq (,$(shell grep "path = $(LIBDIR)" .gitmodules))
 SUBMODULE = true
+endif
 endif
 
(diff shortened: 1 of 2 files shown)


id.mk
```diff
@@ -1,8 +1,10 @@
 ## Identify drafts, types and versions
 
-ifneq (,$(shell git submodule status $(LIBDIR) 2>/dev/null))
+ifneq (,$(wildcard .gitmodules))
+ifneq (,$(shell grep "path = $(LIBDIR)" .gitmodules))
```
martinthomson: You can use `$(and $(wildcard ...),$(shell ... ))` and inline this.  Or you can observe that grep on a non-existent file produces no output to stdout and just do:

ifneq (,$(grep "path *= *$(LIBDIR)" .gitmodules >/dev/null))

Issue #140, opened 2018-02-13
Smaller ToC?


mnot:
One of the things I did in my XSLT was make the both the size and leading of the ToC smaller, to help avoid wrapping / scrolling for more documents. 

WDYT? Happy to give a PR.

mnot:
E.g., see: http://httpwg.org/http-extensions/draft-ietf-httpbis-bcp56bis.html

martinthomson:
Yes, the ToC tends to be too narrow, so smaller text is fine.  As long as you also observe the guidance from Anitra about the multitude of styles.

martinthomson:
You might also consider shaving down the indent for the nested lists.  That's part of the problem.

mnot:
Thoughts?

![screen shot 2018-04-04 at 11 52 21 am](https://user-images.githubusercontent.com/74384/38284458-ad9a2090-37fe-11e8-87e8-c9edb53f0b1d.png)

martinthomson:
I think that we could have slightly bigger font, and cut down the spacing to compensate (line spacing and indentation).

mnot:
If we make it bigger, it's going to be too-similar-but-still-different from the body text; when you make changes in size, they need to look intentional. (I played around with this a fair amount).

mnot:
LGTM, thx.

PR #146, opened 2018-04-04
Smaller ToC

mnot:
for #140

### DIFF
--- a/style.css
+++ b/style.css
@@ -26,10 +26,13 @@
     overscroll-behavior: contain;
   }
   ul.toc li, #rfc\.toc li {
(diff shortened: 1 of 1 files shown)


martinthomson:
This doesn't work with the HTML produced by xml2rfc sadly.

mnot:
Can you give me a sample or point me to one? That way I can test against both. (I've created a local samples/ directory; something to consider?)

martinthomson:
See QUIC.

martinthomson:
Yep, looks good.  I've tweaked further, and simplified a lot.  Let me know what you think of the new look.

PR #148, opened 2018-04-04
Smaller TOC, plus some other tweaks

martinthomson:
This is a bigger change, but I think that it works.  Screenshots to follow.

### DIFF
--- a/style.css
+++ b/style.css
@@ -26,10 +26,13 @@
     overscroll-behavior: contain;
   }
   ul.toc li, #rfc\.toc li {
-    font-size: 15px;
+    font-size: 12px;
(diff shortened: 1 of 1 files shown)


martinthomson:
https://screenshots.firefox.com/WIIoZRM6N1VJmjHW/null is the QUIC repo.  Notice that the headings are much bigger here.  Not much to be done about that sadly.  All else looks fine.

https://screenshots.firefox.com/2Jn31VdM7ruS1nIf/null is the HTTP repo.  The XSLT is better at giving us styles to hook into for headings and all that business.

martinthomson:
@mnot ^^

PR #149, opened 2018-04-04
Tidy up font sizes

mnot:
- abstract a bit bigger, to contract better with surrounding text
- make headings smaller on xml2rfc, to match xslt
- remove small differences in sizes


martinthomson:
That looked awful on windows, until I realized that the font weights were off.  Also, I have a better strategy for making header sizes consistent, see mnot/i-d-template#1.

mnot:
LGTM.

PR #151, opened 2018-04-05
Mobile

mnot:
Depends on adding this to HTML for both XSLT and xml2rfc:

``` html
<meta name="viewport" content="initial-scale=1.0">
```

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,11 +1,3 @@
-@viewport {
-  zoom: 1.0;
-  width: extend-to-zoom;
-}
-@-ms-viewport {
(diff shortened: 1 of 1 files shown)


martinthomson:
Interesting.  The spec says that `@viewport { zoom: 1.0; }` is equivalent to that meta tag.  Maybe we're off the beaten path a little.

mnot:
I think it's just that most browsers don't support @viewport, so they weren't doing anything.

mnot:
And, note requirements upon HTML above.

PR #153, opened 2018-04-12
BSD sed (e.g., OSX) doesn't like this at all.

mnot:


martinthomson:
Standards.  Who needs 'em.

PR #154, opened 2018-04-19
make tags: create tags file to facilitate navigation

dtikhonov:
New utility gen-tags.pl creates file "tags" which can be used by vim
(and, presumably, emacs) users to jump to anchor definitions.  A tag
entry is also generated for each Markdown header.


martinthomson:
This is a little too editor-specific for me to accept this.

I would consider adding a hook that included ~/.i-d-template.mk or something similar if you wanted to include your own targets.

PR #156, opened 2018-05-29
Make the HTML narrower

martinthomson:
This doesn't quite get down to 90 characters per line.  I got most of the way through the "c" on the fourth alphabet, so it's not ideal, but it's still an improvement over previous.  On Mac, Helvetica Neue is slightly wider, so it only gets a partial pixel of the "c", so all win there.

A narrower page allows the TOC to be shown on the side earlier.  It could be made bigger instead, but I think that the size we have is OK.

@mnot, WDYT?

### DIFF
--- a/style.css
+++ b/style.css
@@ -5,7 +5,7 @@ body {
   font-size-adjust: 0.5;
   line-height: 24px;
   margin: 75px auto;
-  max-width: 724px;
+  max-width: 624px;
   padding: 0 5px;
 }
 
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -73,7 +73,7 @@ table.header td, table#rfc\.headerblock td {
 }
 
 samp, tt, code, pre {
-  font: 11pt consolas, monospace;
+  font: 15px Consolas, monospace;
```
mnot: 15 is WAY too big (at least on OSX). 12 looks good, I wouldn't go bigger.

mnot: Oh, wait *px*. Right. That's fine.

PR #160, opened 2018-06-19
Allow things named RFC

ekr:
@martinthomson: this makes things at least sort of work for RFCs.


ekr:
Well, at least it builds, it doesn't make ghpages correctly.

martinthomson:
I'm not seeing any problems with building gh-pages on my machine.

PR #178, opened 2018-12-20
Clean GHPAGES_EXTRA from the target dir.

jyasskin:
https://wicg.github.io/webpackage/loading.html hasn't updated since August, and
https://travis-ci.org/WICG/webpackage/builds/470676744#L1095 seems to indicate it's because `git clone` ran after `loading.html` was built, which led to `/tmp/ghpages5583/loading.html` being newer than its source file, even though it ought to be older.

I don't know what changed in August to make this happen, it only sometimes reproduces locally, and I haven't tested the change on TravisCI, but it seems in line with the intent of the rest of the changed line.

### DIFF
--- a/ghpages.mk
+++ b/ghpages.mk
@@ -90,8 +90,8 @@ cleanup-ghpages: $(GHPAGES_ROOT)
 
 # Clean up contents of target directory
 	@if [ -d $(GHPAGES_TARGET) ]; then \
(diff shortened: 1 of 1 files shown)


martinthomson:
I was baffled by this for a bit myself, but this is right.  The temporary directory is freshly checked out and so the files will have newer modification times than the newly built files.

Thanks for the patch - and apologies for putting you through the process of debugging.

PR #187, opened 2019-04-28
Style: `samp`, `tt`, and `code` in abstracts.

mikewest:
The `samp, tt, code, pre, span.tt` rule overrides the font size specified in the`#rfc\.abstract+p` rule, which looks odd indeed when one of those elements is included in a document's abstract. This patch increases the specificity of the abstract rule to ensure that such elements render correctly.


martinthomson:
I've never really seen those used in abstracts before.  Thanks for being the first.  And fixing it.

PR #189, opened 2019-05-22
Add an extra line to work around weirdness in addstyle.sed

### LINKED ISSUE #188: lib/addstyle.sed breaks if the document body is on one line
If a draft is so boring as to not have any any artwork at all, the `htmltmp` file ends up with the `</style>` on the final line. That breaks the saving and restoring hack. The script never starts another cycle to restore the saved line.
https://github.com/martinthomson/i-d-template/blob/master/addstyle.sed#L19

martinthomson:
Closes #188.

@davidben does this work for you?

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -84,8 +84,9 @@ else
 endif
 
 %.html: %.htmltmp $(LIBDIR)/addstyle.sed $(LIBDIR)/style.css
+# Note that the extra `echo` here is to work around a limitation in addstyle.sed
 ifneq (,$(if $(ENABLE_RIBBON),$(CI_REPO_FULL),))
-	 (cat $<;echo) | sed -f $(LIBDIR)/addstyle.sed -f $(LIBDIR)/addribbon.sed | \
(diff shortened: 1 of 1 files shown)


main.mk
```diff
@@ -85,10 +85,10 @@ endif
 
 %.html: %.htmltmp $(LIBDIR)/addstyle.sed $(LIBDIR)/style.css
 ifneq (,$(if $(ENABLE_RIBBON),$(CI_REPO_FULL),))
-	sed -f $(LIBDIR)/addstyle.sed -f $(LIBDIR)/addribbon.sed $< | \
+	 (cat $<;echo) | sed -f $(LIBDIR)/addstyle.sed -f $(LIBDIR)/addribbon.sed | \
```
davidben: Worth a comment here as to why the `echo`?

davidben:
Does this get picked up by the http-extensions repo automatically or do I need to wait something to update on that end?

martinthomson:
This should be picked up automatically.  Though a local copy needs to be updated.

Issue #186, opened 2019-04-23
Be able to run pyang


richsalz:
In xml2rfc v3, YANG code is marked with
```
<sourcecode type="yang">
...
</sourcecode>
```

It would be nice to strip those out and feed them through ```pyang --ietf``` for syntax-checks, and perhaps also ```pyang -f yang --keep-comments``` And maybe put ```max-line-length 69``` somewhere.

martinthomson:
I've left this sort of linting to other tools.  The Makefile for the QUIC drafts [adds a lint](https://github.com/quicwg/base-drafts/blob/7fbcdb0e7c66bbe9193a96cd21519ebd7a02644a/Makefile#L24) that looks for line length, for instance.  To that, I would use the native inclusion functions of the tool, whatever that is, and add a lint that uses `pyang` for those files.  Something like:

```xml
<sourcecode type="yang">
<xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="file.yang" parse="text"/>
</sourcecode>
```

...and:

```make
lint::
        pyang --ietf *.yang
```

That said, it's probably good to integrate this sort of check into the tool proper.  It's fairly easy to extract the relevant pieces using xmllint.

martinthomson:
Of note here is a new [option for people who want more tools in CI images](https://github.com/martinthomson/i-d-template/blob/main/doc/LINTING.md#using-the-mega-image-in-ci).  pyang is probably small enough to include in the core image; but if it is not, then that option exists.

mcr:
> ```
> <sourcecode type="yang">
> ```

You can't include the yang directly like this, as it needs CDATA, I think.

mcr:
At: https://github.com/anima-wg/voucher/runs/5436309463?check_suite_focus=true I notice that the pyang in the circleCI container is too old to understand --keep-comments.
@martinthomson I don't remember how you want us to report problems like this.

martinthomson:
@mcr, the image I'm using is just 3 days old (as of your comment), so if pyang was updated in the intervening time to include support for that argument, that's expected.  Do I need to rebuild the image for a new version?

mcr:
%pyang --help | grep keep-comment
  --keep-comments       Pyang will not discard comments;
%pyang --version                 
pyang 2.5.0

Could I be using something from git, rather than released?
% which pyang
/home/mcr/.local/bin/pyang
% ls -l ~/.local/bin/pyang
[...]

PR #193, opened 2019-09-16
add yanglint feature

GrumpyOldTroll:
Hi,

I wrote a "make yanglint" feature that's been useful for me when working on yang-related drafts.  I cleaned it up some, and I'd like to get it integrated into your repo so it'll end up merged in the future, if possible.

It greps markdown for some strings and injects the external yang file contents for `YANG-MODULE <module>`, or the pyang tree for `YANG-TREE <module>`, or an example json data file contents for `YANG-DATA <module> <file.json>` during a regular make.

If you run `make yanglint` or set a `VALIDATE_YANG=1` environment variable, it will run yanglint and pyang --lint to report errors, using those same search strings.

There's an example repo that uses this feature here:
https://github.com/GrumpyOldTroll/ietf-taps-yang

I looked through the issues, and this is related to the request in #186, but I think approaches it a little differently than that request seems to suggest:

This feature only works on markdown, as it stands, and I don't think it breaks anything if your draft doesn't use the feature.

Would you be willing to add this, or to give some feedback about next steps to make it acceptable if it's not there yet?

Thanks and regards,
Jake

PS: I hereby dedicate the code for this to the public domain--if you want something more formal lmk


yang-fetch-imports.sh
```diff
@@ -0,0 +1,102 @@
...
+
+if [ ! -d modules/ ]; then
+  mkdir -p modules/
+fi
+
+before=$(ls modules/ | wc -l)
+after=${before}
+last=$((before-1))
+
+echo "checking for module dependencies (downloading into modules/...)"
```
martinthomson: This should write into a dot directory that is added to `.gitignore` by default, but if an environment variable is set, then it should choose that directory instead.  Though the main modules need to be in source control, these extra dependencies should just be downloaded into a scratch space.  Having the environment variable means that you can share a cache between drafts just by setting a value.

yang-inject.sh
```diff
@@ -0,0 +1,74 @@
...
+had_error=0
+
+target_xml="$(echo ${input_md} | sed -e 's/.md$/.xml/')"
+for ((i=0; i < ${#data_jsons[@]}; i++)); do
+  if [ ! -f "${data_jsons[i]}" ]; then
+    echo "no example data file ${data_jsons[i]}"
+    had_error=1
+  fi
+done
+for mod in ${all_modules}; do
```
martinthomson: These files could use better quoting hygiene.  Bash arrays can help here.  Then you would do "${all_modules[@]}".

yang-check.sh
```diff
@@ -0,0 +1,177 @@
...
+      for loc in ${locals}; do
+        if [ "${loc}" = "${missing}.yang" ]; then
+          already=1
+        fi
+      done
+      if [ "${already}" != "0" ]; then
+        break
+      fi
+      locals="${locals} ${missing}.yang"
+      missing=$( ( yanglint -f json -D -V -p . -p modules ${mod} 2>&1 || true ) | \
```
martinthomson: This looks like code repetition.  Use a function perhaps.

GrumpyOldTroll:
Thanks for the feedback, this is very helpful.

I missed the existence of `{::include fname}` for kramdown-rfc2629 and `<{{include fname}}` for mmark.  I'll refactor based on those, address your comments, and resubmit.

Just to make sure I understand: I think the main design suggestion is to do a build operation that covers *.yang, instead of extracting explicit references from the .md.  I think that makes sense.

yang-check.sh
```diff
@@ -0,0 +1,177 @@
...
+    missing=$( ( pyang --verbose --path modules:. ${mod} 2>&1 || true ) | \
+      grep "error: module" | \
+      sed -e 's/.*: error: module "\([^"]*\)" not found in search path.*/\1/')
+    if [ "$missing" != "" ]; then
+      if ! which rsync 2>&1 > /dev/null ; then
+        echo "error in $0: rsync required; please install"
+      fi
+
+      for mis in ${missing}; do
+        bash -e -x -c "rsync -cvz rsync.iana.org::assignments/yang-parameters/${mis}*.yang modules/"
```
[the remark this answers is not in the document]
GrumpyOldTroll: I can use rsync directly, sure.  I was doing this mainly so that when rsync errors, the call that caused it will be visible, but the quoting is a very good point and I think it should echo instead, thanks.

yang-check.sh
```diff
@@ -0,0 +1,177 @@
...
+fi
+
+fatal=0
+if ! which pyang 2>&1 > /dev/null ; then
+  echo "error in $0: pyang required; please install\n(pip install pyang)"
+  fatal=1
+fi
+
+if ! which yanglint 2>&1 > /dev/null ; then
+  echo "error in $0: yanglint required; please install libyang:"
```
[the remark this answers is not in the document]
GrumpyOldTroll: That seems a bit rude and maybe requires privileges to add something in PATH.  I guess maybe install into an environment variable location when it's missing and look for it there?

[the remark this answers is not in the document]
martinthomson: I wasn't suggesting that you put this in /usr/local or anything like that.  But you can download and compile the tool and run that copy.  You can see an example of that sort of process in https://github.com/w3c/push-api/blob/gh-pages/Makefile

martinthomson:
> Just to make sure I understand: I think the main design suggestion is to do a build operation that covers *.yang, instead of extracting explicit references from the .md. I think that makes sense.

👍

This doesn't do anything for checking examples, but maybe you can address that in a separate patch.  It seems like having examples inline is of more use than the module definitions themselves, so we might have better justification for extracting those, but I'm sure that we can work something out. Pulling examples from the XML might be the best, but that requires some interesting tweaks to the order of operations, so I think we might want to start with the big bang stuff.

yanglint.mk
```diff
@@ -0,0 +1,16 @@
...
+	  if ! $(LIBDIR)/yang-check.sh $< ; then \
+	    echo "$(LIBDIR)/yang-check.sh $< failed" ; \
+	    exit 1 ; \
+	  fi ; \
+	else \
+		echo "(no yang imported in $<)" ; \
+	fi
+
+yanglint: $(CHECK_TARGETS)
+
```
Andreykarp251187, line 16:
```suggestion
2019-10-30 12#244 
```
[RFC 2119 - Key.txt](https://github.com/martinthomson/i-d-template/files/5325359/RFC.2119.-.Key.txt)

```

Comment (Andreykarp251187) [reply yanglint.mk:16]:
```suggestion
2019-10-30 12#244 
```

martinthomson:
@billwuqin, that's certainly a viable strategy.  This pull request is a bit old, which means that it isn't necessarily how you would do it today.

You might do better with integrating the lint more tightly as I describe [in the documentation](https://github.com/martinthomson/i-d-template/blob/main/doc/TOOLS.md#linting-tools).  For something like `pyang-lint`, you can get automated installation as described on the same page or in [the sourcecode](https://github.com/martinthomson/i-d-template/blob/main/deps.mk).

That means that you would only need to create a `requirements.txt` and add a rule approximately like:

```make
yang-files := $(wildcard *.yang)
yang-file-markers := $(patsubst %.yang,.%.yang-lint,$(yang-files))
.PHONY: pyang-lint
pyang-lint: $(pyang-lint-files)
.%.yang-lint: %.yang $(DEPS_FILES)
        pyang-lint $< | tee $@
lint:: pyang-lint
```

(I haven't debugged the above, but I hope that it is clear enough.)

billwuqin:
> @billwuqin, that's certainly a viable strategy. This pull request is a bit old, which means that it isn't necessarily how you would do it today.
> 
> You might do better with integrating the lint more tightly as I describe [in the documentation](https://github.com/martinthomson/i-d-template/blob/main/doc/TOOLS.md#linting-tools). For something like `pyang-lint`, you can get automated installation as described on the same page or in [the sourcecode](https://github.com/martinthomson/i-d-template/blob/main/deps.mk).
> 
> That means that you would only need to create a `requirements.txt` and add a rule approximately like:
> 
> ```makefile
> yang-files := $(wildcard *.yang)
> yang-file-markers := $(patsubst %.yang,.%.yang-lint,$(yang-files))
> .PHONY: pyang-lint
> pyang-lint: $(pyang-lint-files)
> .%.yang-lint: %.yang $(DEPS_FILES)
>         pyang-lint $< | tee $@
> lint:: pyang-lint
> ```
> 
> (I haven't debugged the above, but I hope that it is clear enough.)

Thanks, Martin.

PR #195, opened 2019-10-07
Add an $(XML2RFCV3) variable to opt into the xml2rfc v3 path.

jyasskin:
The other tools aren't totally ready for this, but this'll make it easier to find their bugs.

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -55,6 +55,8 @@ endif
 %.xml: %.org
 	$(oxtradoc) -m outline-to-xml -n "$@" $< > $@
 
+XML2RFCV3OPTION := $(if $(V3),--v3)
+
 XSLTDIR ?= $(LIBDIR)/rfc2629xslt
 ifeq (true,$(USE_XSLT))
 $(LIBDIR)/rfc2629.xslt: $(XSLTDIR)/rfc2629.xslt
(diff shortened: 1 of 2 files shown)


main.mk
```diff
@@ -55,6 +55,8 @@ endif
 %.xml: %.org
 	$(oxtradoc) -m outline-to-xml -n "$@" $< > $@
 
+XML2RFCV3OPTION := $(if $(V3),--v3)
```
martinthomson: I've tended to use the following form:

```suggestion
ifeq (true,$(XML2RFCV3))
XML2RFCV3 := --v3
endif
```

It's more verbose, but it makes the check - and the variable - more explicit.

main.mk
```diff
@@ -55,7 +55,11 @@ endif
 %.xml: %.org
 	$(oxtradoc) -m outline-to-xml -n "$@" $< > $@
 
-XML2RFCV3OPTION := $(if $(V3),--v3)
+ifeq (true,$(XML2RFCV3))
+XML2RFCV3 := --v3
```
martinthomson: This is a little odd, overwriting the value of the tested variable, but it does work.

jyasskin: Oh, I thought that's what you'd suggested in https://github.com/martinthomson/i-d-template/pull/195#discussion_r332904139. I don't have any particular opinion about the best way to do this. :)

martinthomson: I must have typed the same thing twice :/  I'll sort it out.

jyasskin:
Thanks! (8578e0572a00625d4972d3d1cc4ce3081dc50f4a)

PR #198, opened 2019-10-14
Also build with `--raw`

felixhandte:
The `--raw` output from `xml2rfc` is sometimes more useful than the regular `--txt` build (when diffing, for example). This very simple PR adds it to the versions that are built by default. Tested with `make`, `make draft-foo.raw.txt`, and `make clean`.

I've found this personally useful. If you feel that this is not a valuable contribution, feel free to close. My feelings will not be hurt. :)

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -1,5 +1,5 @@
 .PHONY: latest
-latest:: txt raw html
+latest:: txt html
 
 .DELETE_ON_ERROR:
 
--- a/main.mk
+++ b/main.mk
@@ -1,5 +1,5 @@
 .PHONY: latest
-latest:: txt html
+latest:: txt raw html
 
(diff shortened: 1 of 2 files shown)


main.mk
```diff
@@ -1,5 +1,5 @@
 .PHONY: latest
-latest:: txt html
```
martinthomson: I'd prefer it if you leave this line alone.  Building raw by default is an option you can enable by adding

```make
latest:: raw
```

to your `Makefile`.

martinthomson:
Thanks!

martinthomson:
BTW, the xml2rfcv3 output won't be paginated, so you might try XML2RFCV3=true instead (#195).

PR #201, opened 2019-10-31
adds github actions support

jam01:


martinthomson:
Actions support is now first class.

PR #203, opened 2019-11-17
Produce v3 XML

martinthomson:
This runs the v3 conversion tool when processing markdown.  New styles go with that.


main.mk
```diff
@@ -55,7 +60,7 @@ ifdef REFCACHEDIR
 endif
 
 %.xml: %.org
-	$(oxtradoc) -m outline-to-xml -n "$@" $< > $@
+	$(oxtradoc) -m outline-to-xml -n "$@" $< | $(xml2rfc) --v2v3 /dev/stdin -o $@
 
 ifeq (true,$(XML2RFCV3))
 XML2RFCV3OPTION := --v3
```
jyasskin: Now that you're unconditionally generating V3 XML, the `--v3` option appears not to do anything. Should we revert #195?

martinthomson: Thanks!  I've done that manually.

PR #206, opened 2020-03-05
Script to archive repo state

MikeBishop:
Updated version of #194, using the GitHub v4 API and incorporating suggestions from that review.  It's technically longer, but most of that length is pieces of GraphQL queries.

Note that this takes your suggestion to create a custom file layout rather than strictly dumping the API output, but in the process breaks your issues.{html,js} code.  I haven't attempted to fix that so far.

### DIFF
--- a/archive_repo.py
+++ b/archive_repo.py
@@ -557,12 +557,12 @@ if args.refFile:
 
             if fileIsValid:
                 lastSuccess = dp.parse(raw_reference["timestamp"])
-                for issue in raw_reference["issues"]:
-                    issue_ref[issue["number"]] = issue
+                issue_ref = dict(
+                    [(issue["number"], issue) for issue in raw_reference["issues"]]
+                )
                 if "pulls" in raw_reference:
                     ref_is_issues_only = False
-                    for pr in raw_reference["pulls"]:
-                        pr_ref[pr["number"]] = pr
+                    pr_ref = dict([(pr["number"], pr) for pr in raw_reference["pulls"]])
     except:
         warnings.warn("Unable to read input file; proceeding without it")
         pass
@@ -609,9 +609,9 @@ while get_more_issues:
         issue["labels"] = [label["name"] for label in issue["labels"]["nodes"]]
 
         # Delete the old instance; add this instance
-        if issue["number"] in issue_ref.keys():
-            del issue_ref[issue["number"]]
-        issue_ref[issue["number"]] = issue
+        if number in issue_ref.keys():
+            del issue_ref[number]
+        issue_ref[number] = issue
         just_copy_old_issues = False
 
     get_more_issues = issues["pageInfo"]["hasNextPage"]
--- a/archive_repo.py
+++ b/archive_repo.py
@@ -111,7 +111,6 @@ gql_Paged = """
             endCursor
             hasNextPage
         }
-        totalCount
 """
 
 # Issues
--- a/archive_repo.py
+++ b/archive_repo.py
@@ -15,9 +15,7 @@ import warnings
 parser = argparse.ArgumentParser(description="Archive repo issues and PRs.")
 parser.add_argument("repo", help="GitHub repo to archive (e.g. quicwg/base-drafts)")
 parser.add_argument("githubToken", help="GitHub OAuth token")
-parser.add_argument(
-    "outFile", default="archive_repo.json", nargs="?", help="destination for output"
-)
+parser.add_argument("outFile", default=None, nargs="?", help="destination for output")
 parser.add_argument(
     "--reference oldfile",
     dest="refFile",
@@ -99,6 +97,23 @@ fragment commentFields on Comment {
 }
 """
 
+gql_RateLimit = """
+fragment rateLimit on Query {
+    rateLimit {
+        remaining
+        resetAt
+    }
+}
+"""
+
+gql_Paged = """
+        pageInfo {
+            endCursor
+            hasNextPage
+        }
+        totalCount
+"""
+
 # Issues
 
 gql_Issue_Fields = (
@@ -118,11 +133,9 @@ fragment issueFields on Issue {
             ...author
             ...commentFields
         }
-        pageInfo {
-            endCursor
-            hasNextPage
-        }
-        totalCount
+        """
+    + gql_Paged
+    + """
     }
 }
 """
@@ -134,20 +147,16 @@ fragment issueFields on Issue {
 gql_Issues_Query = (
     """
       nodes { ...issueFields }
-      pageInfo {
-        endCursor
-        hasNextPage
-      }
-      totalCount
+        """
+    + gql_Paged
+    + """
     }
   }
-  rateLimit {
-    remaining
-    resetAt
-  }
+  ... rateLimit
 }
 """
     + gql_Issue_Fields
+    + gql_RateLimit
 )
 
 gql_AllIssues_First = (
@@ -196,22 +205,18 @@ query($id: ID!, $cursor: String!){
                     ...author
                     ...commentFields
                 }
-                pageInfo {
-                    endCursor
-                    hasNextPage
-                }
-                totalCount
+        """
+    + gql_Paged
+    + """
             }
         }
     }
-    rateLimit {
-        remaining
-        resetAt
-    }
+    ...rateLimit
 }
 """
     + gql_Comment_Fields
     + gql_AuthorFields
+    + gql_RateLimit
 )
 
 # Pull Requests
@@ -229,11 +234,9 @@ fragment reviewFields on PullRequestReview {
             originalPosition
             ...commentFields
         }
-        pageInfo {
-            endCursor
-            hasNextPage
-        }
-        totalCount
+        """
+    + gql_Paged
+    + """
     }
 }
 """
@@ -259,21 +262,17 @@ fragment prFields on PullRequest {
         nodes {
             ...author
             ...commentFields }
-        pageInfo {
-            endCursor
-            hasNextPage
-        }
-        totalCount
+        """
+    + gql_Paged
+    + """
     }
     reviews(first: 50) {
         nodes {
             ...reviewFields
         }
-        pageInfo {
-            endCursor
-            hasNextPage
-        }
-        totalCount
+        """
+    + gql_Paged
+    + """
     }
 }
 """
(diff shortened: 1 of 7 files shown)


archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+
+    bodyjson = {"query": re.sub("\s+", " ", query).strip()}
+    if variables:
+        bodyjson["variables"] = variables
+    body = json.dumps(bodyjson)
+
+    if not args.quiet:
+        output = f"Submitting query for {display} with "
+        output += str(variables) if variables else "no parameters"
+        print(output)
```
martinthomson: Maybe this is worth defining:

```python
def eprint(*str, **kwargs):
    print(*str, file=sys.stderr, **kwargs)

if args.quiet:
    def log(**kwargs):
        pass
else:
    def log(*str, **kwargs):
        eprint(*str, **kwargs)
```

Writing progress messages to stderr and writing the archive to stdout seems like the nicest approach here, absent a `-o` argument.

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+                    "GitHub API rate-limited; waiting for"
+                    + str(time_to_sleep)
+                    + "seconds"
+                )
+                time.sleep(time_to_sleep)
+                continue
+            else:
+                print(e.request.url)
+                print(e.request.headers)
+                print(e.response.headers)
```
martinthomson: These might be worth making conditional on `not args.quiet`, i.e., use the log function.
```suggestion
                log(e.request.url)
                log(e.request.headers)
                log(e.response.headers)
```

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+                print(e.response.content)
```
martinthomson:
```suggestion
                eprint(e.response.content)
```

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+                if attempt != 2:
+                    delay = 10 * (num_retries + 1)
+                    print(f"Retrying ({num_retries}) after {delay} seconds...")
```
martinthomson:
```suggestion
                    eprint(f"Retrying ({num_retries}) after {delay} seconds...")
```

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+    result = response.json()
+
+    if "data" in result.keys() and result["data"] is not None:
+        if "rateLimit" in result["data"]:
+            last_request_limit = result["data"]["rateLimit"]["remaining"]
+            if last_request_limit < 2:
+                # We're about to be rate-limited; STALL
+                reset = dp.parse(result["data"]["rateLimit"]["resetAt"])
+                time_to_sleep = int(reset - datetime.datetime.now()) + 1
+                print(
```
martinthomson:
```suggestion
                eprint(
```

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+            fileIsValid = True
+            for element in ("magic", "timestamp", "issues", "repo", "version"):
+                if element not in raw_reference:
+                    fileIsValid = False
+                    break
+            if not fileIsValid or raw_reference["magic"] != "B8n2c@e8kvfx":
+                warnings.warn("Input file does not appear to be generated by this tool")
+                fileIsValid = False
+
+            if fileIsValid and raw_reference["version"] != "1.0":
```
martinthomson: Why include version AND magic both?

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+        "timestamp": now.isoformat(),
+        "version": "1.0",
+        "repo": args.repo,
+        "issues": [issue for (id, issue) in sorted(issue_ref.items())],
+    }
+    if not args.issuesOnly:
+        output["pulls"] = [pr for (id, pr) in sorted(pr_ref.items())]
+
+    with open(args.outFile, "w") as output_file:
+        json.dump(output, output_file, indent=4)
```
martinthomson:
```suggestion
        json.dump(output, output_file, indent=2)
```

Less side-scrolling :)

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+)
+
+gql_PR_ReviewComments_Query = (
+    """
+query($id: ID!, $cursor: String!){
+    node(id: $id) {
+        ...on PullRequestReview {
+            comments(first: 50, after:$cursor) {
+                nodes {
+                    originalPosition
```
martinthomson: Are you sure that we want `originalPosition` and not `position`?

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+        hasNextPage
+      }
+      totalCount
+    }
+  }
+  rateLimit {
+    remaining
+    resetAt
+  }
+}
```
martinthomson: Unbalanced braces do make this slightly harder to read.

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+    mergedBy { ...userFields }
+    comments(first: 100) {
+        nodes {
+            ...author
+            ...commentFields }
+        pageInfo {
+            endCursor
+            hasNextPage
+        }
+        totalCount
```
martinthomson: It might pay to factor out this block into a gql_Paged variable.

Separately, I need to check that you really need totalCount for this.

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+                    hasNextPage
+                }
+                totalCount
+            }
+        }
+    }
+    rateLimit {
+        remaining
+        resetAt
+    }
```
martinthomson: It would be awesome if this could be added to every query automatically, but if it can't then a variable is probably better.

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+
+import os
+import sys
+import json
+import requests
+import argparse
+from datetime import datetime, timezone
+import time
+import re
+import dateutil.parser as dp
```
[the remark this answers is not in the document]
MikeBishop: Note that I added installing it to the Docker file.  `apt install python3-dateutil`, I believe.

archive_repo.py
```diff
@@ -0,0 +1,667 @@
...
+
+if not args.githubToken and "GH_TOKEN" in os.environ.keys():
+    args.githubToken = os.environ["GH_TOKEN"]
+
+if args.repo[-1] == "/":
+    args.repo = args.repo[:-1]
+
+API_headers = {"user-agent": "MartinThomson/i-d-template/archive_repo.py"}
+
+API_headers["authorization"] = "bearer " + args.githubToken
```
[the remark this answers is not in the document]
MikeBishop: Good catch.  In the previous version, the token was optional, and I didn't collapse these when I removed the `if:`

archive_repo.py
MikeBishop: I tried, but couldn't get it to work as a fragment because fragments have to declare what type they're on, and these fields are perversely not part of any interface I could find in the documentation.  Python3 formatted strings choke on all the braces in a GraphQL expression, so it's not simple to just insert as an inline variable.  I could get it in with a sentinel and a `.replace()` call, but that doesn't really feel any cleaner than just repeating it.

If you figure out how, though, I'd be happy to take the change.

No, I don't use totalCount for anything; I was going to, but GitHub helpfully added the hasNextPage since I last took a stab at this.  While I thought there might be value in having it to cross-check, we can drop it with no loss of functionality (besides not having to count the JSON objects to display the statistic when traversing the file).

archive_repo.py
MikeBishop: The alternative is to repeat this piece several times.  I figured proper indentation could make this the lesser of two evils.

archive_repo.py
MikeBishop: I'll see if it's possible to pull into a fragment.

archive_repo.py
MikeBishop: Frankly, I'm not sufficiently clear on the difference.  What I'd really like is a line number in the commit, rather than the line number in the diff (original or not), but I don't see that in the API.

MikeBishop:
I do read the names of the labels on issues and pull requests (see `gql_LabelFields`); easy enough to grab the description as well, though it's so often the empty string I didn't think it was worth the space.

archive_repo.py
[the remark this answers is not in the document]
martinthomson: Ack.  I thought that I removed this comment when I found that.  Sorry for the noise.

archive_repo.py
martinthomson: This is just a string though, so you could do:

```python
gql_Paged = """
pageInfo {
    endCursor
    hasNextPage
}
"""

gql_PullRequest_Fields = (
    """
fragment prFields on PullRequest {
    number
    id
    title
    url
    state
    ...labels
    ...author
    ...commentFields
    closedAt
    mergedAt
    mergedBy { ...userFields }
    comments(first: 100) {
        nodes {
            ...author
            ...commentFields }
"""
+ gql_Paged 
"""
    }
    reviews(first: 50) {
        nodes {
            ...reviewFields
        }
"""
+ gql_Paged 
"""
    }
}
"""
    + gql_LabelFields
    + gql_Review_Fields
)
```

Or you could get fancy and do (with more formatting):

```python
def paged(name, fields, first=10, cursor=False):
    name + ("(first=%d" % first) + (", after:$cursor" if cursor else "") + ") { nodes { " + fields + "pageInfo { endCursor, hasNextPage } } }"
```

archive_repo.py
martinthomson: It's just one line.

archive_repo.py
martinthomson: Yeah, I saw that; it's not very nice.  I suspect that this is a feature that GH is stuck with.  It's probably why you can't comment on lines that aren't in the diff as well.  But maybe they will eventually be able to fix it, who knows.

martinthomson:
For labels, perhaps the way to do this is pull labels into another top-level structure with the necessary data, then collapse the overly complex `labels: { nodes: [] }` into just `labels: ["name", "name"]`.  That's post-processing I'm happy to take on when I try to work out how to do the offline display piece.

archive_repo.py
MikeBishop: Then I think we're talking about different things.  Do you have a suggested change?

MikeBishop:
@martinthomson, the labels are now collected into a top-level element and collapsed to arrays of names in each issue/PR.

archive_repo.py
MikeBishop: I briefly sketched out something like the fancier approach, but there are too many different variations for me to think it's worthwhile, what with issue filters and parameter types.  By the time you pass in all the info it needs to construct the query, you're passing in the query.  The simpler version is there, though, with a "gql_Paged" between two block strings.

archive_repo.py
```diff
@@ -0,0 +1,729 @@
...
+query($id: ID!, $cursor: String!){
+    node(id: $id) {
+        ...on Issue {
+            comments(first:100, after:$cursor) {
+                nodes {
+                    ...author
+                    ...commentFields
+                }
+        """
+    + gql_Paged
```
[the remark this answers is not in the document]
MikeBishop: I tried -- the curly braces made Python go crazy.  You could use multiple {{}}s, I think, but I don't think that improves readability.

PR #207, opened 2020-03-14
Trim history of gh-pages

MikeBishop:
If the oldest commit on gh-pages is more than 60 days old, drop all commits from gh-pages which are more than 30 days old.  Note that this involves a force-push of gh-pages back to origin, but that should generally be okay; no one is likely to be basing their own commits off of gh-pages.

Hopefully this should permit GitHub to garbage collect a ton of old and very large state, reducing the size of the QUIC repo in the near future.  (And, of course, the document versions being dropped can be rebuilt from the git history of the main branch.)

_(Note: the Behave tests failed on this commit, but the failing test is for `make ghissues` and shouldn't be impacted by this change.  Also, the test succeeds on my local branch.  I'm not quite sure what to make of the failure.)_

### DIFF
--- a/ghpages.mk
+++ b/ghpages.mk
@@ -78,11 +78,10 @@ cleanup-ghpages: $(GHPAGES_ROOT)
 	done;
 
 # Drop old gh-pages commits (keep 30-60 days of history)
-	@CUTOFF_30=$$(($$(date '+%s')-2592000)); \
-	CUTOFF_60=$$(($$(date '+%s')-5184000)); \
-	ROOT=$$(git -C $(GHPAGES_TARGET) rev-list --max-parents=0 gh-pages); \
-	if [ `git -C $(GHPAGES_ROOT) show -s --format=%ct $$ROOT` -lt $$CUTOFF_60 ]; then \
-	  git -C $(GHPAGES_ROOT) replace --graft `git rev-list --min-age=$$CUTOFF_30 --max-count=1 gh-pages`; \
+	@KEEP=$$((`date '+%s'`-2592000)); CUTOFF=$$((`date '+%s'`-5184000)); \
+	ROOT=`git -C $(GHPAGES_TARGET) rev-list --max-parents=0 gh-pages`; \
+	if [ `git -C $(GHPAGES_ROOT) show -s --format=%ct $$ROOT` -lt $$CUTOFF ]; then \
+	  git -C $(GHPAGES_ROOT) replace --graft `git rev-list --min-age=$$KEEP --max-count=1 gh-pages` && \
 	  git -C $(GHPAGES_ROOT) filter-branch gh-pages; \
 	fi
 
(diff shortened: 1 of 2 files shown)


ghpages.mk
```diff
@@ -77,6 +77,15 @@ cleanup-ghpages: $(GHPAGES_ROOT)
 	  git remote prune $$remote; \
 	done;
 
+# Drop old gh-pages commits (keep 30-60 days of history)
+	@CUTOFF_30=$$(($$(date '+%s')-2592000)); \
+	CUTOFF_60=$$(($$(date '+%s')-5184000)); \
+	ROOT=$$(git -C $(GHPAGES_TARGET) rev-list --max-parents=0 gh-pages); \
+	if [ `git -C $(GHPAGES_ROOT) show -s --format=%ct $$ROOT` -lt $$CUTOFF_60 ]; then \
+	  git -C $(GHPAGES_ROOT) replace --graft `git rev-list --min-age=$$CUTOFF_30 --max-count=1 gh-pages`; \
+	  git -C $(GHPAGES_ROOT) filter-branch gh-pages; \
```
martinthomson:
```suggestion
	@KEEP=$$((`date '+%s'`-2592000)); CUTOFF=$$((`date '+%s'`-5184000)); \
	ROOT=`git -C $(GHPAGES_TARGET) rev-list --max-parents=0 gh-pages`; \
	if [ `git -C $(GHPAGES_ROOT) show -s --format=%ct $$ROOT` -lt $$CUTOFF ]; then \
	  git -C $(GHPAGES_ROOT) replace --graft `git rev-list --min-age=$$KEEP --max-count=1 gh-pages` && \
	  git -C $(GHPAGES_ROOT) filter-branch gh-pages; \
```

The renaming and switch to backticks is merely an attempt to get this consistent.  The && is important.  The cleanup happens before we add new commits and if the repo has been idle for the last 30 days, I don't want to generate an error.

ghpages.mk
```diff
@@ -103,13 +111,13 @@ ghpages: cleanup-ghpages $(GHPAGES_ALL)
 	  git -C $(GHPAGES_ROOT) $(CI_AUTHOR) commit -m "Script updating gh-pages from $(shell git rev-parse --short HEAD). [ci skip]"; fi
 ifeq (true,$(PUSH_GHPAGES))
 ifneq (,$(if $(CI_HAS_WRITE_KEY),1,$(if $(GH_TOKEN),,1)))
-	git -C $(GHPAGES_ROOT) push https://github.com/$(GITHUB_REPO_FULL) gh-pages
+	git -C $(GHPAGES_ROOT) push https://github.com/$(GITHUB_REPO_FULL) gh-pages --force
```
martinthomson:
```suggestion
	git -C $(GHPAGES_ROOT) push -f https://github.com/$(GITHUB_REPO_FULL) gh-pages
```

Better to be consistent with flag positioning, and I think that everyone knows what `-f` does.

ghpages.mk
```diff
@@ -103,13 +111,13 @@ ghpages: cleanup-ghpages $(GHPAGES_ALL)
...
 else
-	@echo git -C $(GHPAGES_ROOT) push -q https://github.com/$(GITHUB_REPO_FULL) gh-pages
-	@git -C $(GHPAGES_ROOT) push -q https://$(GH_TOKEN)@github.com/$(GITHUB_REPO_FULL) gh-pages >/dev/null 2>&1
+	@echo git -C $(GHPAGES_ROOT) push -q https://github.com/$(GITHUB_REPO_FULL) gh-pages --force
+	@git -C $(GHPAGES_ROOT) push -q https://$(GH_TOKEN)@github.com/$(GITHUB_REPO_FULL) gh-pages --force >/dev/null 2>&1
```
martinthomson:
```suggestion
	@echo git -C $(GHPAGES_ROOT) push -qf https://github.com/$(GITHUB_REPO_FULL) gh-pages
	@git -C $(GHPAGES_ROOT) push -qf https://$(GH_TOKEN)@github.com/$(GITHUB_REPO_FULL) gh-pages >/dev/null 2>&1
```

ghpages.mk
```diff
@@ -103,13 +111,13 @@ ghpages: cleanup-ghpages $(GHPAGES_ALL)
...
 endif
 else
-	git -C $(GHPAGES_ROOT) push origin gh-pages
+	git -C $(GHPAGES_ROOT) push origin gh-pages --force
```
martinthomson:
```suggestion
	git -C $(GHPAGES_ROOT) push -f origin gh-pages
```

martinthomson:
The final `--aggressive` size was 238M, so maybe it did have some effect.  Though it could just be the result of better compression; this repo has a lot of redundancy and it did take a fabulously long time.

MikeBishop:
You may also need to clear your reflog; the old commits are still referenced by the history of HEAD, so they can't be deleted until that expires.  However, GitHub is pretty aggressive about garbage collecting from what I've read, so I'd expect the reduction in size on the server side to take effect a few hours after the commits get orphaned.

ghpages.mk
MikeBishop: I _think_ this should already work on a 30-day-idle repo, since it chooses the first commit older than 30 days as the one to keep.  If there's been no activity for 30 days, then however old the most recent commit is, that's the commit that will be chosen as the new beginning-of-time for the gh-pages branch, and the new state will be added as the child.

MikeBishop:
Yep -- specifically, the `reflogExpireUnreachable` variable defaults to 30 days, meaning unreachable commits that are still in a branch's reflog will be kept by gc.  Also, until you push, the commits can't be pruned because the remote repo's ref still points into the old tree.

Unreachable commits aren't transferred, even if they still exist in the local store.  A fresh checkout of QUIC is 198MB.  After doing this on my fork of the QUIC repo, a fresh checkout is 28MB.  However, doing `git gc` on my old copy of the repo makes it balloon, and I'm not clear why.  (Possibly I've invalidated so many packed objects that it decides to unpack everything?)

PR #209, opened 2020-03-16
My iteration on the archive script

martinthomson:
@MikeBishop WDYT?  Do you want to see a PR on your branch?

### DIFF
--- a/archive_repo.py
+++ b/archive_repo.py
@@ -557,12 +557,12 @@ if args.refFile:
 
             if fileIsValid:
                 lastSuccess = dp.parse(raw_reference["timestamp"])
-                for issue in raw_reference["issues"]:
-                    issue_ref[issue["number"]] = issue
+                issue_ref = dict(
+                    [(issue["number"], issue) for issue in raw_reference["issues"]]
+                )
                 if "pulls" in raw_reference:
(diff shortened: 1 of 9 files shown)


martinthomson:
The email address came through empty in my testing.  Were you able to get an email address somehow?

I wasn't confident in removing the `userFields` fragment though.  I assume that you would just have `author { login }` where you wanted that information.

MikeBishop:
You can select a public email address in your Github profile; it's empty for anyone who hasn't. It's not universally empty, or universally filled.

I'd love to have an "if not empty," but GraphQL doesn't have the concept of filtering on values to decide whether to return something.

martinthomson:
I see.  In that case, I think I'll limit this to the login name.  I realize that the email thing might help in the sense that it has meaning in IETF-land, but I don't think we need to avoid using the user profile for information related to users.  For instance, I had to go looking for another way to display an avatar to get the HTML viewer working.  This could save information about every user to the archive, such as email address and avatars, but that can wait.  Right now, I think the priority is getting the information displayed correctly.

MikeBishop:
Perhaps a similar treatment to labels - capture just the login in issues, then gather the list of emails, avatars, etc. in a separate top-level list.

PR #212, opened 2020-03-17
Rate-limiting code

MikeBishop:
Should work, but not tested directly because I don't have a rate-limited token.


martinthomson:
Let's give this a go then.

PR #214, opened 2020-03-17
Fix the ratelimit fix

MikeBishop:
My fix changed the scope of the loop and left something unset.  🤦‍♂️


MikeBishop:
I tried this (with your GH_TOKEN) from the Circle instance.  It still hiccups if GitHub just starts 403-ing the requests without explanation, but there's not much we can do in that case except try again later.

PR #222, opened 2020-04-07
Handle deleted users when archiving.

jyasskin:
This broke https://travis-ci.org/github/WICG/webpackage/jobs/671886121, and you
can see an example issue at https://github.com/WICG/webpackage/issues/128.


martinthomson:
Thanks Jeffrey.  I didn't know that was even possible (my expectation was that the login name would remain). I wonder if we should be using a different key.

PR #223, opened 2020-04-23
PR fields and upgrade in archive

MikeBishop:
Adds a few extra fields to PRs:
- baseRepository { nameWithOwner }
- baseRefName
- baseRefOid
- headRepository { nameWithOwner }
- headRefName
- headRefOid
- mergeCommit { oid }

Since it's the first time we've changed the query, it also creates a mechanism for upgrading old files to new versions.

The issue and PR retrieval are pulled into functions.  These functions take a fragment with the fields that need to be fetched; the full run will get all fields, while upgrade passes in the fields that have changed.  The functions are also more tolerant of things not being present on the returned object.

There's a dictionary of old versions and the fields that need to get changed to bring them up to a newer version.  We could always bring them to the current version, or chain them, either way will work.  After reading the file, if it's old, it calls the functions to fetch the newly-added fields on all the old objects.

I'm a little split on whether to include the upgrade or just indicate whether you need to dump the old objects and download fresh (i.e. this PR without the last two commits).  You wind up visiting all the old objects anyway to get the new fields; having an upgrade path for the reference saves bandwidth on the retrieval and probably consumes ratelimit less quickly, but doesn't actually save any calls.


archive_repo.py
```diff
@@ -503,20 +511,22 @@ def followPagination(node, key, query, display):
...
     "Collapse something in the form of { x: nodes [] }"
-    thing[key] = thing[key]["nodes"]
+    if key in thing:
+        thing[key] = thing[key]["nodes"]
 
 
-def collapse_to_login(thing, key):
-    """Collapse something in the form of { x: {login: 'login'} }
+def collapse_map(thing, key, name):
+    """Collapse something in the form of { x: {$name: 'login'} }
```
martinthomson, line 525:
```suggestion
    """Collapse something in the form of { x: {$name: $value} } into {x: $value}
```

archive_repo.py
```diff
@@ -683,34 +663,134 @@ def log(*str, **kwargs):
...
+(owner, repo) = args.repo.split("/", 1)
+
+## Read in the reference files, if any
+issue_ref = dict()
+pr_ref = dict()
+lastSuccess = None
+ref_is_issues_only = True
+
+if args.refFile:
+    try:
```
martinthomson, line 739: It seems like this might be worth pushing into a new function.  Or two, called `load_reference` and `update_reference`.  The first can do the simple loading, ignorant of any upgrade logic.  The second can turn the loaded reference into a reference of the current version.

archive_repo.py
```diff
@@ -535,106 +545,73 @@ def log(*str, **kwargs):
...
 
-    get_more_issues = issues["pageInfo"]["hasNextPage"]
-    issue_cursor = issues["pageInfo"]["endCursor"]
+        get_more_issues = issues["pageInfo"]["hasNextPage"]
+        issue_cursor = issues["pageInfo"]["endCursor"]
 
-## Similar process with PRs, except they don't have a filter
-if not args.issuesOnly:
 
+def getPRs(fields=gql_PullRequest_Fields, updateOld=False):
```
martinthomson, line 611: It might be better to pass the reference in as a parameter here.

archive_repo.py
```diff
@@ -656,11 +633,14 @@ def log(*str, **kwargs):
             number = pr["number"]
 
             # Since we can't filter, check if we already have this one.
-            if number in pr_ref:
-                ref_updatedAt = dp.parse(pr_ref[number]["updatedAt"])
-                dl_updatedAt = dp.parse(pr["updatedAt"])
-                if ref_updatedAt >= dl_updatedAt:
-                    continue
+            if not updateOld:
+                if number in pr_ref:
```
martinthomson, line 637: Don't you need to declare `global pr_ref` in order to get access to global state?

MikeBishop: Technically, no; `global` declares that I'm going to be reassigning a global variable within this scope instead of creating a new variable with the same name.  Accessing globals and interacting with the objects they point at doesn't require a declaration.  Once pr_ref is assigned, it's never reassigned; everything is mutations of the object.

MikeBishop:
The trouble with consistency is that we'd expect all changes to modify the output.  If we take a stance that it's additive-only, we could check that everything the last version of master produces also appears in the newly-submitted version, perhaps?

martinthomson:
Let's make the change then iterate.

Issue #246, opened 2020-10-14
build-index.sh


mnot:
I'm starting to use a jekyll remote template to pull together the look and feel of how httpwg repos are presented, and it's going well.

However, the `index.html` created by i-d-template poses two problems:

1. It conflicts with taking up the README file as the default for the repo, which is sometimes desirable.
2. It's in HTML, not markdown, so it's more difficult to style (e.g., adding a navigation bar).

I think this could be addressed by:

1. Making the name of the file configurable
2. Allowing optional markdown output instead of HTML

@martinthomson if you're amenable I suspect you'd be more efficient at (1); I can take a stab at (2), although hooking in the control mechanism might be better done by you. WDYT?

mnot:
Oh, this would imply that the branch written to (and presumably the directory) would be configurable too.

martinthomson:
@mnot - is this now satisfactory?

mnot:
Hmm. I've set `MD_INDEX := true`, but it still generates `index.html`, which overrides the markdown for index views.

martinthomson:
Ah, something got lost somewhere.  Try `INDEX_FORMAT := md` instead.

mnot:
Seems to work fine, thanks! I've got it going for `http-extensions` and `http2-spec` now.

@LPardue if you want https://quicwg.org/base-drafts/ to be prettier (see e.g., https://httpwg.org/http-extensions/ ) try this; you'll need to have a theme repo (like https://github.com/httpwg/theme) and point to it in `_config.yml`.

PR #247, opened 2020-10-15
rough in build-index-md

mnot:
For #246


mnot:
@martinthomson I think this is roughed in, doubtless it needs some polish.

PR #248, opened 2020-10-19
Markdown indexes for gh-pages

martinthomson:
@mnot, does this work for you?

### DIFF
--- a/build-index-md.sh
+++ b/build-index-md.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+
+# Usage: $0 [dir] [gh-user] [gh-repo] > index.md
+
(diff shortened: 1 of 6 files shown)


mnot:
Yes, it seems to, thanks.

It would be super-nice to be able to include a header and/or footer from a file in `main` at some point (both in HTML and Markdown). 

A more general version of that (at least for Markdown) would be to give control over the `layout` in the YAML front matter.

martinthomson:
I don't know which YAML front matter you refer to there.  There are potentially multiple drafts from which this can draw, but at the point that this runs the drafts have all been compiled down to HTML and text.  What might be possible is to specify a template that contains the structure of the index, maybe with a placeholder for where all the content goes.  Future work perhaps.

PR #249, opened 2020-10-26
Add documentation on linting

martinthomson:
This is rough.

### DIFF
--- a/doc/LINTING.md
+++ b/doc/LINTING.md
@@ -56,7 +56,7 @@ modified target in `.github/workflows/ghpages.yml` and
     - name: "Build Drafts"
       uses: martinthomson/i-d-template@v1
       with:
-	make: latest sf-rfc-validate=/root/.local/bin/sf-rfc-validate
+        make: latest sf-rfc-validate=/root/.local/bin/sf-rfc-validate
 ```
 
 Alternatively, you can attempt to detect that you are running in CI and adjust
--- a/doc/LINTING.md
+++ b/doc/LINTING.md
@@ -48,7 +48,9 @@ when running in CI.
 
 One way to do this is to ensure that the binary is identified when running
 `make`.  Modifying `.github/workflows/*.yml` or `.circleci/config.yml` to
-include additional arguments to make should do this:
+include additional arguments to make should do this.  The following is a
+modified target in `.github/workflows/ghpages.yml` and
+`.github/workflows/publish.yml`:
 
 ```yaml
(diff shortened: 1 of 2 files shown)


doc/LINTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+.PHONY: sf-lint-install
+sf-lint-install:
+	@hash sf-rfc-validate 2>/dev/null || pip3 install --user sf-rfc-validate
+```
+
+Note that for things like python, the location that the file is installed to
+might not be on the path, so you will need to ensure that you modify the path
+when running in CI.  The best way to do this is to ensure that the binary is
+identified when running `make`.  Modifying `.github/workflows/*.yml` or
```
mnot: for GitHub, is this just `ghpages.yaml`?

doc/LINTING.md
```diff
@@ -0,0 +1,57 @@
...
+`.circleci/config.yml` to include additional arguments to make should do this:
+
+```yaml
+    - name: "Build Drafts"
```
mnot, line 54: Build Drafts?

doc/LINTING.md
martinthomson: That depends on whether you want the linting to occur before you publish.  So I would probably add something there as well.  Technically, I think that only the targets in `ghpages.yaml` runs the lint, but you might want to add the same check to `publish.yml` also so that you don't publish without first linting.  I'll add a note.

doc/LINTING.md
martinthomson: This is the name of the target in `ghpages.yml`, that's all.

mnot: Yes - before it was something else. Can close.

doc/LINTING.md
```diff
@@ -0,0 +1,72 @@
...
+`make`.  Modifying `.github/workflows/*.yml` or `.circleci/config.yml` to
+include additional arguments to make should do this.  The following is a
+modified target in `.github/workflows/ghpages.yml` and
+`.github/workflows/publish.yml`:
+
+```yaml
+    - name: "Build Drafts"
+      uses: martinthomson/i-d-template@v1
+      with:
+	make: latest sf-rfc-validate=/root/.local/bin/sf-rfc-validate
```
martinthomson, line 59:
```suggestion
        make: latest sf-rfc-validate=/root/.local/bin/sf-rfc-validate
```

PR #250, opened 2020-11-04
add CDDL validation and CBOR diagnostic tools

thomas-fossati:
This adds `cddl(1)` and `cbor-diag(1)` to the CI docker image toolchain.

It allows validating inlined CDDL and checking CBOR examples against CDDL.

An example usage is the RATS EAT spec:

* https://github.com/ietf-rats-wg/eat/tree/master/cddl
* https://github.com/ietf-rats-wg/eat/tree/master/Makefile

It looks to me like it could be generally useful?

cheers!

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,6 +33,8 @@ RUN apt-get update && \
  && rm -rf /var/lib/apt/lists/* /var/log/dpkg.log \
  && apt-get autoremove -y && apt-get clean -y
 
+RUN gem install cddl cbor-diag
+
 ENV USER idci
 ENV LOGNAME $USER
 ENV HOSTNAME $USER


docker/Dockerfile
```diff
@@ -33,6 +33,8 @@ RUN apt-get update && \
  && rm -rf /var/lib/apt/lists/* /var/log/dpkg.log \
  && apt-get autoremove -y && apt-get clean -y
 
+RUN gem install cddl cbor-diag
+
```
martinthomson, line 37: There is a gem install step further down that these could be added to, but see the meta-comment for more.

martinthomson (commented):
I maintain two docker images out of this repo, this only adds to one of those.

However, I've been working on a process for tools like this - which aren't used by every document - to be integrated into the process.  We could gather a list of tools like this and put them all in these images, but that would bloat the images quite a bit.  Instead, I've put together [a guide](https://github.com/martinthomson/i-d-template/blob/main/doc/LINTING.md) on how tools can be integrated on an as-needed basis.  I think that this is better.

Over time, it's possible that widespread use of particular tools means that it will be easier to have them integrated into these images, but I would prefer to have people develop experience with adding additional validation tools.

thomas-fossati:
> However, I've been working on a process for tools like this - which aren't used by every document - to be integrated into the process. We could gather a list of tools like this and put them all in these images, but that would bloat the images quite a bit. Instead, I've put together [a guide](https://github.com/martinthomson/i-d-template/blob/main/doc/LINTING.md) on how tools can be integrated on an as-needed basis. I think that this is better.

Thanks, I hadn't seen this, my bad.  Looks pretty neat.

> Over time, it's possible that widespread use of particular tools means that it will be easier to have them integrated into these images, but I would prefer to have people develop experience with adding additional validation tools.

Agree.  The image that includes these is 319M -- vs 310M of the original, a ~3% increase for just a couple of tools that not everyone is going to use is clearly sub-optimal.

PR #259, opened 2021-02-02
Install tex2svg and asciitex

larseggert:
This will hopefully allow https://github.com/NTAP/rfc8312bis to build, which uses the new `~~~ math` artwork that kramdown-rfc2629 now supports.

### DIFF
--- a/action/Dockerfile
+++ b/action/Dockerfile
@@ -74,7 +74,5 @@ RUN set -e; v="0.31.7"; sha="a2a21eccc2d7485f1807ec0e8f234d77dc622e5843c0bddd3ff
     mv "$tmp" "$target"; \
     chmod 755 "$target"
 
-RUN ls -lta /usr/local/bin
-
 COPY entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
--- a/action/Dockerfile
+++ b/action/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update && \
     libxml2-utils \
     locales \
     make \
+    npm \
     python-is-python3 \
     python3-certifi \
     python3-chardet \
@@ -63,6 +64,17 @@ RUN set -e; v="2.2.9"; sha="b8c501be4118f79a294b658d963ec8c91595e94aea69c8efd0ea
 
 RUN pip3 install --compile xml2rfc
 RUN gem install --no-doc kramdown-rfc2629
+
+RUN npm install -g mathjax-node-cli
+
+RUN set -e; v="0.31.7"; sha="a2a21eccc2d7485f1807ec0e8f234d77dc622e5843c0bddd3ff63b4d80d4a735"; \
+    tmp=$(mktemp -t "asciitexXXXXXX"); target="${BINDIR:-/usr/local/bin}/asciitex"; \
+    curl -sSLf "https://github.com/larseggert/asciiTeX/releases/download/asciiTeX-${v}/asciitex" -o "$tmp"; \
+    [ $(sha256sum -b "$tmp" | cut -d ' ' -f 1 -) = "$sha" ]; \
+    mv "$tmp" "$target"; \
+    chmod 755 "$target"
+
+RUN ls -lta /usr/local/bin
 
 COPY entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update && \
     libxml2-utils \
     locales \
     make \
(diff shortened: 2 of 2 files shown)


martinthomson:
This docker file is only used by CircleCI; if you want GH actions, you need to update the other too.

Just checking now about the size difference.  I really didn't want to have to install node and npm here; large images add significantly to the running time of CI.

martinthomson:
sooo...many...node...packages.  Building the image takes a very long time with all those node modules.  And that's just before the 172 extra for mathjax-node-cli.

Here's the delta (my reference image is old, but it won't have changed much):

```
lars-math                                       latest              3fa2750122ce        5 seconds ago       530MB
martinthomson/i-d-template                      latest              b53f78b1eb11        8 weeks ago         318MB
```

That's a pretty steep tax.  I'll need to think about it.

larseggert:
I made the change to the other Dockerfile, just so that the PR is at least complete.

Yes, the size issue sucks. I hate `node`.

larseggert:
Have you considered using Alpine instead of Ubuntu for your build image? It is much (~150MB) smaller by default.

martinthomson:
Would it really make that much of a difference?  The main problem is that there are lots of dependencies from the stuff this already uses (git alone is terrible, but ruby and python aren't small either).

I'd consider a pull request that made the change though.

martinthomson:
I have an alpine image working and it is ~150M smaller, but I don't think that I want to use that reduction to justify an increase.

The problem here is that there is so much advantage in having the tools in the base image.  So it makes sense to have different images for people who need the math. But in order to use a different image, you need to change where your configuration file points.  For GitHub, that also means defining an entirely new action as well.  You could do that in your fork of the repository (you can copy the actions I use to create the image, probably skipping dockerhub uploads).  But the end result is pretty poor because you now have a continual rebasing problem to work through.

How about we try this:  Take a look at the alpine branch on this repo and see if you can make a new docker/math subdirectory with a dockerfile that uses the one from docker/action as a baseline.  Then, I will create a new branch for action.yml and we can change the file to point to this new image, tag that as `math` or something, and we'll see how that works.  Then, you can just tweak your configuration to use the `@math` version of the action.

That might be the way to do this without adding 200M to the size of the image everyone else uses.

larseggert:
Is the docker image you build from action/Dockerfile available? Because then I could base the `math` image on it (and basically only require downloading the delta layers) rather than making a fully separate image that shares much of the same underlying Dockerfile?

larseggert:
Are the tags `martinthomson/i-d-template-action` for the GitHub actions image and `martinthomson/i-d-template` for the CircleCI one?

larseggert:
For now, I've just duplicated most of `action/Dockerfile` for the `math/Dockerfile` in #260.

larseggert:
FWIW, if I install all the dependencies that `kramdown-rfc2629` might possibly want, the image is at 1.5GB...

martinthomson:
Hi @larseggert, I've published a package that contains the updated image.  This isn't mainline yet, because I want to test it.  Can you modify your .github/workflows on the rfc8312bis repo to use `martinthomson/i-d-template@v1m` instead of `@v1` throughout and see how that works out?  You might need to remove the additions to the `Makefile` that install stuff using `apt-get`, because that won't work with these alpine-based images.

The image is ~340M, which isn't ideal, but I now have a "maxi-sized" image branch that can be used for other things.  I don't think that a super-maxi image of 1.5G is something I ever want though.  I guess that rules out making Rust available :).

larseggert:
I tried that , but `*** [math:] sh: tex2svg: not found`. See https://github.com/NTAP/rfc8312bis/pull/40/checks?check_run_id=1854225262#step:6:16

martinthomson:
Oops, a PATH issue.  I forgot to add `-g` to `npm install`.  Try again?

(BTW, that's a LOT of node modules.)

larseggert:
I did "Re-run all jobs" but am getting the same error. The image is still called `martinthomson/i-d-template@v1m`, right?

martinthomson:
Thanks for checking.  I think that this is a genuine problem with what kramdown-rfc2629 and the rest of the stuff you installed pull down.  Now we just need to find svgcheck.

martinthomson:
I've just pushed a change to include svgcheck in the main image (it's small).  That might take a little while to build, but it should be good to try again in a little while.

larseggert:
Now it worked: https://github.com/NTAP/rfc8312bis/actions/runs/547794430

martinthomson:
I think that we can close this now.

larseggert:
So should I use `martinthomson/i-d-template@v1m` now going forward?

martinthomson:
If you want to use the image with the extra tools installed, yeah.

PR #266, opened 2021-02-11
Make setup-note.sh work on macOS

DavidSchinazi:
Bash case operators were added in Bash 4, and macOS runs a
version of Bash from the Paleolithic. This uses tr instead to
support older versions of Bash.


martinthomson:
Really.  Mac bash strikes again.

PR #269, opened 2021-03-10
Update SETUP.md

eckelcu:
I am in the process of getting myself setup as described here (thanks!). 
I am not sure if this was a typo or if things changed, but I needed to do "brew install libxslt" instead of "brew install libxslt1".

$ brew install libxslt1
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
Error: No available formula or cask with the name "libxslt1".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
==> Searching taps on GitHub...
Error: No formulae found in taps.

This is with MacOS. I did not try on other platforms.

### DIFF
--- a/doc/SETUP.md
+++ b/doc/SETUP.md
@@ -60,7 +60,7 @@ $ sudo pip3 install xml2rfc
 ```
 
 xml2rfc might need development versions of [libxml2](http://xmlsoft.org/) and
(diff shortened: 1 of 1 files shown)


martinthomson:
Thanks Charles.

PR #270, opened 2021-03-13
Migrate off tools.ietf.org

LPardue:
may as well avoid some indirection


martinthomson (approved):
Thanks.  I've done this myself and it's good.  (Though DT still lacks some important features, this isn't it.)

PR #274, opened 2021-04-07
Remove archive_repo; rely on pip

MikeBishop:
I'm unsure how best to test this; suggestions welcome.  I deliberately removed python3-dateutil to ensure we're exercising pip's dependency install process; I don't think it's needed except as an import in the archive script.


martinthomson:
Thanks for doing this Mike.  I think that this is the right time.

It might pay to add to the setup documentation that includes the installation of this tool.  Most people won't use that, but it is also very small and can be installed alongside xml2rfc.

PR #275, opened 2021-04-09
Allow fetching default branch without credentials

DavidSchinazi:
On public GitHub repositories, the API call for the default
branch succeeds even without credentials. This PR changes
default-branch.py to attempt the API call even if there are
no credentials present.


martinthomson (approved):
Thanks David.  I seem to recall issues on CI systems resulted in this being unwise - as the anonymous call pretty much always hits the rate limit.  As the calling code doesn't depend on this succeeding (it already fails noisily) I think it's OK to try this.

PR #278, opened 2021-04-21
Sketch in support for per-draft issue links

mnot:
Currently only supports Markdown drafts.


mnot:
See https://httpwg.org/http-extensions/ for an example

build-index.sh
```diff
@@ -168,6 +174,12 @@ function list_dir() {
         fi
 	      diff=$(rfcdiff "https://tools.ietf.org/id/${file}.txt" "$this_githubio")
         td "$(a "$diff" 'diff with last submission' diff "$file")"
+        this_issue_label=$(./lib/extract-frontmatter.py ${file}.md github-issue-label)
```
martinthomson, line 177: How about a fallback based on the draft name:

```
echo "${file}" | cut -d- -f 4- -
```

That is, for "draft-ietf-wg-short-name", this would produce "short-name" as a default.

build-index.sh
```diff
@@ -168,6 +174,12 @@ function list_dir() {
...
+        if [[ "$this_issue_label" ]]; then
```
martinthomson, line 178: You don't need to go looking for a label if `[[ "${#files[@]}" -eq 1 ]]`.  The issues link at the top of the page is enough in that case.

martinthomson (commented):
Seems broadly reasonable.  You are relying on markdown though.  Yes, the source file is correct if there is a markdown file (and it is kramdown), but that assumes a lot.

It might be better to change the way this works to take all the input file names as additional arguments, rather than using `find`.  Then you know what the inputs are (And can force them to be in a common format (probably XML).  Of course, that leaves the problem of ensuring that the metadata survives and can be read from the XML.  Maybe that is enough hard work that you don't want to even contemplate that.

build-index.sh
mnot: That's somewhat likely to result in links that don't work; I'd rather have a positive assertion of the label from the author.

mnot:
Hm. That seems like a more substantial re-architecture.

mnot:
XML now supported (via PIs).

This is going to need PyYaml to run; is that just a matter of adding it to `docker/action/Dockerfile`, or is something else necessary?

martinthomson:
Yeah, docker/action/Dockerfile should cover it.

build-index.sh
```diff
@@ -168,6 +174,16 @@ function list_dir() {
         fi
 	      diff=$(rfcdiff "https://tools.ietf.org/id/${file}.txt" "$this_githubio")
         td "$(a "$diff" 'diff with last submission' diff "$file")"
+        if [[ "${#files[@]}" -eq 1 ]]; then
+            td ""
+        else
+            this_issue_label=$(./lib/extract-frontmatter.py ${file}.md github-issue-label)
```
martinthomson, line 180:
```suggestion
            this_issue_label=$(./lib/extract-frontmatter.py $(ls ${file}.{md,xml}) github-issue-label)
```

Or something like that; though fixing this up is probably something I can sort out.

extract-frontmatter.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
martinthomson, line 1: This is really extract-metadata rather than extract-frontmatter, isn't it?

build-index.sh
```diff
@@ -27,6 +27,12 @@ function githubio() {
     echo "https://${user}.github.io/${repo}/${d#$default_branch/}${2}.txt"
 }
 
+function githubcom() {
+    d="${1%/}/"
```
martinthomson, line 31: You don't *need* the trailing slash added here, so I wouldn't bother with it.

build-index.sh
mnot: Hmm. Either that or move the logic into the script (i.e., just hand it the base name). Preferences?

mnot:
Any other changes necessary here?

martinthomson:
I think that we're mostly there.  I realize now that the XML change is going to break...  Leave it with me.  I have a rough idea of how to fix it, but need to find time in between stints of QUIC AUTH48.

mnot:
so...

PR #281, opened 2021-05-13
Update latest manual workflow

martinduke:


martinthomson:
Hey, this is the automated process (the first on the list).  I provide the manual thing as a backup (some people won't have CI running), but these steps are already there [at the top](https://github.com/martinthomson/i-d-template/blob/main/doc/SUBMITTING.md#automated-process).

martinduke:
Uh, I am clearly missing something obvious, but 'git push --tags' is not mentioned in the unicorn example, just in the bug reference.

martinthomson:
`git push --tags` is a shorthand.  There are problems with CircleCI that means that you really need to do `git push origin $TAG` one by one instead.  But `git push --tags` is what I gave you in chat because that works with GitHub Actions and it's shorter.  And I was typing on my phone.

martinduke:
Sorry, I'll close it.

Broadly, this documentation is not meeting the threshold of allowing users to treat it as a magic script without knowing anything about anything, which is where I'm at.

Nevertheless, thank you very much for these tools -- they are outstanding.

PR #284, opened 2021-06-08
Issue labels

martinthomson:
@mnot, can you review my changes here?  I've tested it and it works for me.  I've made a few tweaks so that it's more robust.

Also, I notice that kramdown complains about the github-issue-label attribute being unused.  It might pay to ask Carsten for a special space for metadata so that non-kramdown tools can add attributes without generating warnings.

Closes #278.


mnot (approved):
Looks entirely reasonable to me. Let's see how it goes.

martinthomson:
I pre-warmed the image, so this should just work...

mnot:
Works, thanks!

PR #286, opened 2021-06-09
Non greedy

### LINKED ISSUE #285: Pre-commit always fails for a draft whose name includes "rfc"
The change to the sed regex in c3e60fa1c2748cf60da5201706d4256e0b38ca3e is apparently greedy -- in a filename like `draft-foo-rfc9000bis.md`, it produces the target `rfc9000bis.md` instead.  Assuming this file doesn't exist, the pre-commit check always fails.  (If it does exist, the pre-commit check is building the wrong document.)

MikeBishop:
Fixes #285, but I'm not sure exactly how portable this is.  Worth testing on some other platforms before merging, perhaps.

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -23,7 +23,7 @@ function abort() {
 trap abort ERR
 trap cleanup EXIT
 
-files=($(git status --porcelain draft-* rfc* | sed '/^[[:space:]]*[MARCU]/{s/[[:space:]]*[MARCU][[:space:]]*\(draft-\|rfc\)/\1/;p;};d' | sort))
+files=($(git status --porcelain draft-* rfc* | sed '/^[MARCU]/{s/.* \(draft-\|rfc\)/\1/;p;};d' | sort))
 for f in "${files[@]}"; do
     tmp="${f%.*}"-tmp$$."${f##*.}"
     tmpfiles+=("$tmp")
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -23,7 +23,7 @@ function abort() {
 trap abort ERR
 trap cleanup EXIT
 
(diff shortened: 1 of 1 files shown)


pre-commit.sh
```diff
@@ -23,7 +23,7 @@ function abort() {
 trap abort ERR
 trap cleanup EXIT
 
-files=($(git status --porcelain draft-* rfc* | sed '/^[MARCU]/{s/.*\(draft-\|rfc\)/\1/;p;};d' | sort))
+files=($(git status --porcelain draft-* rfc* | sed '/^[[:space:]]*[MARCU]/{s/[[:space:]]*[MARCU][[:space:]]*\(draft-\|rfc\)/\1/;p;};d' | sort))
```
martinthomson, line 26: How about a simpler approach:

```suggestion
files=($(git status --porcelain draft-* rfc* | sed '/^[MARCU]/{s/.* \(draft-\|rfc\)/\1/;p;};d' | sort))
```

The man-page for `git status` indicates that the fields are separated by a single space, so this should work without making this too complicated.

MikeBishop: Yes, that should work.  The non-porcelain output has more whitespace there, but this should work for porcelain.

MikeBishop:
Is this waiting on anything to merge?

martinthomson:
Sorry, I missed the notice.

laurencelundblade:
Hi,

I don't understand this SW stack very well, but circumstantial evidence I see is that this change broke something. I use this for https://github.com/ietf-rats-wg/eat. It builds OK on MacOS with this, but the Circle CI build is broken. It also looks like Circle CI is broken here too.

martinthomson:
See #293 where we're tracking this one.  I am equally in the dark.

PR #292, opened 2021-06-16
Run tools inside the i-d-toolchain Docker image

larseggert:
If USE_DOCKER is defined, this PR will have the Makefile run all commands inside an [i-d-toolchain](https://github.com/larseggert/i-d-toolchain) Docker container.

The image for this container can be updated via the "update-docker" `make` target.

This means that the only thing that needs to be installed locally is `make`.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -2,11 +2,23 @@
 # All are assumed to be on the path, but you can override these
 # in the environment, or command line.
 
+ifneq (,$(USE_DOCKER))
+XML2RFC_REFCACHEDIR := /id/.cache/xml2rfc
+KRAMDOWN_REFCACHEDIR := $(XML2RFC_REFCACHEDIR)
+IMAGE := ghcr.io/larseggert/i-d-toolchain:latest
+DOCKER := docker run \
+		--env XML2RFC_REFCACHEDIR=${XML2RFC_REFCACHEDIR} \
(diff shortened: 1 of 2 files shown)


martinthomson (commented):
It might be easier to provide a different approach entirely: 

By providing a different Makefile, you can have `make` run inside the docker image.

For example:

```make
%:
    docker run .... make "$@"
```

With something like that, you don't even need the mess that clones this repo into `./lib`.  That can happen inside the image (where the image already contains this repo).

larseggert:
Yeah, this actually already works without a different Makefile. The downside is that the command to run is quite cumbersome, i.e.,
```
docker run --env XML2RFC_REFCACHEDIR=/id/.cache/xml2rfc --env KRAMDOWN_REFCACHEDIR=/id/.cache/xml2rfc --volume /Users/lars/Documents/IETF/md-for-authors:/id:delegated --interactive --cap-add=SYS_ADMIN ghcr.io/larseggert/i-d-toolchain:latest make
```

larseggert:
Good idea. I'll add `Makefile.docker`, and the invocation could then become `make -f Makefile.docker` instead of just `make`, which is manageable.

larseggert:
I added an [i-d-toolchain bash script](https://github.com/larseggert/i-d-toolchain/blob/main/i-d-toolchain) to https://github.com/larseggert/i-d-toolchain.

PR #298, opened 2021-07-04
Update REPO.md

mcr:
This name no longer works, needs to be replaced with some other instruction?
I created this as a pull request, but it is more of an issue, but I wanted to point exactly where the instruction is wrong.


msigoo33 (approved):

martinthomson:
Thanks for reporting the issue @mcr.  I've a fix for this that isn't quite as elegant as a direct link, but the function of that link is obviously long gone.

PR #299, opened 2021-07-07
added pyang to image

mcr:
add pyang to big image


martinthomson (commented):
ISTM that this is a candidate for the main repo given how many people are likely to use it.  I just tested this out and it adds just 4Mb to the image size.  It might add less if it were installed as part of the same layer as xml2rfc.  That's not nothing, but I'd say that's worthwhile.

What do you think @mcr?

mcr:
If it works in main, great.  I don't care how I get pyang, as long as I get it :-)

martinthomson:
OK, problem solved then.

Issue #279, opened 2021-05-08
Add kgt (Kate's grammar tool) to docker


cabo:
https://github.com/cabo/kramdown-rfc2629/commit/ce91004c7278c527b3d85a240ec6186e24115617

cabo:
(Usage example:
https://github.com/core-wg/coral/commit/2dd2ba
)

cabo:
Oh, and what *is* the schedule for updating the docker image?
(I'd like to use kramdown-rfc 1.4.11 as well.)

martinthomson:
It runs twice a month, but I can poke it for you.

martinthomson:
In general, I don't object to useful tools being added as long as they are small (Lars wanted to add 1G, which was too much).  

kgt seems small enough, but as it isn't packaged on alpine, it requires manual steps to install.  That's very tricky.  

The default install doesn't have a C compiler and the build would be considerably more complex if one where to be added (you would have to do a two-stage build where the file is compiled at the first stage in a large image and then copied to the second stage).  

I'm happy to take a pull request; I know how to do the work, so I could easily review a proposed change, but it's not work that I'm able to commit to doing at this stage.

cabo:
I just pushed 1.4.12 :-)

(Maybe twice a month is slightly on the slow side.)

I know nothing about Alpine and its packaging, so I cannot efficiently address that part; maybe I can ask around.

PR #301, opened 2021-07-25
Add kgt to math image

cabo:


cabo:
Close #279

martinthomson:
Thanks Carsten.  This is the right way to do it.

PR #306, opened 2021-09-01
Titles, not filenames

mnot:
Have tested the Python changes, but not the shell adjustments.

### DIFF
--- a/build-index.sh
+++ b/build-index.sh
@@ -186,7 +186,7 @@ function list_dir() {
     table_i "branch-$2"
     for file in "${files[@]}"; do
         dir=$(dirname "$file")
-        title=$("${libdir}/extract-metadata.py" "$i" title)
+        title=$("${libdir}/extract-metadata.py" "$i" abbrev)
         file=$(basename "$file" .txt)
 
         tr_i
--- a/build-index.sh
+++ b/build-index.sh
@@ -186,10 +186,11 @@ function list_dir() {
     table_i "branch-$2"
     for file in "${files[@]}"; do
         dir=$(dirname "$file")
+        title=$("${libdir}/extract-metadata.py" "$i" title)
         file=$(basename "$file" .txt)
 
         tr_i
-        th "${file}"
+        th "${title}"
         td "$(a "$(reldot "$dir")/${file}.html" html html "$file")"
         td "$(a "$(reldot "$dir")/${file}.txt" "plain text" txt "$file")"
         this_githubio=$(githubio "$branch${dir#$root}" "$file")
--- a/extract-metadata.py
+++ b/extract-metadata.py
@@ -20,18 +20,44 @@ def extract_md(filename):
 
 def extract_xml(filename):
     parser = xml.sax.make_parser()
-    handler = PIHandler()
+    handler = MdHandler()
     parser.setContentHandler(handler)
     parser.parse(filename)
-    return handler.pis
+    return handler.md
 
 
-class PIHandler(xml.sax.handler.ContentHandler):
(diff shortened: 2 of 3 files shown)


extract-metadata.py
```diff
@@ -20,18 +20,44 @@ def extract_md(filename):
 
 def extract_xml(filename):
     parser = xml.sax.make_parser()
-    handler = PIHandler()
+    handler = MdHandler()
```
martinthomson, line 23: Yeah, that name is unfortunate.  Perhaps we can avoid "md" as a shorthand for "metadata" when it can (and does) already mean "markdown".

```suggestion
    handler = XmlHandler()
```

extract-metadata.py
```diff
@@ -20,18 +20,44 @@ def extract_md(filename):
...
     parser.setContentHandler(handler)
     parser.parse(filename)
-    return handler.pis
+    return handler.md
```
martinthomson, line 26:
```suggestion
    return handler.metadata
```

Similar naming throughout.

extract-metadata.py
```diff
@@ -44,7 +70,10 @@ def processingInstruction(self, target, data):
         fileext = os.path.splitext(filename)[1]
         extract_func = extract_funcs.get(fileext, lambda a: {})
         frontmatter = extract_func(filename)
-        value = frontmatter.get(target, "")
+        if target == "title" and frontmatter.get("abbrev", None) != None:
+            value = frontmatter["abbrev"]
```
martinthomson, line 74: Wouldn't it make sense to query for abbrev if you want the abbreviated name?

Then abbrev has a special fallback to title, as follows:

```python
if target == "abbrev":
    value = frontmatter.get("abbrev", frontmatter.get("title", ""))
else:
    value = frontmatter.get(target, "")
```

martinthomson (commented):
Seems like a good idea (I've used draft names, but abbreviated titles might be more user-friendly).

build-index.sh
```diff
@@ -186,10 +186,11 @@ function list_dir() {
     table_i "branch-$2"
     for file in "${files[@]}"; do
         dir=$(dirname "$file")
+        title=$("${libdir}/extract-metadata.py" "$i" title)
```
martinthomson, line 189:
```suggestion
        title=$("${libdir}/extract-metadata.py" "$i" abbrev)
```

extract-metadata.py
martinthomson:
```suggestion
            value = frontmatter.get("abbrev", frontmatter.get("title", ""))
```

extract-metadata.py
```diff
@@ -44,7 +70,10 @@ def processingInstruction(self, target, data):
         fileext = os.path.splitext(filename)[1]
         extract_func = extract_funcs.get(fileext, lambda a: {})
         frontmatter = extract_func(filename)
-        value = frontmatter.get(target, "")
+        if target == "title" and frontmatter.get("abbrev", None) != None:
```
martinthomson, line 73: OK, github won't let me do a multi-line suggestion here for some reason, so you get two...

```suggestion
        if target == "abbrev":
```

martinthomson (commented):
I think that this is what you want.

martinthomson:
One thing that might trip us up here is that sometimes (at least in XML) the title can contain newlines.  We might need to do whitespace normalization (I know, ugh).

mnot:
Ah, yeah, I strip it but don't normalise. Hold on...

martinthomson:
Thanks.  I'll take this without the normalization for now.

PR #315, opened 2021-10-19
Use the ietf.org rfcdiff

mnot:
Note that the first URL fed to rfcdiff still leads to tools.ietf.org, because datatracker and www don't have a 'latest version' URL for the text format.


martinthomson:
Thanks.  I had something staged, but forgot to follow up after Robert answered my question about it.

PR #317, opened 2021-11-01
Point RFC URLs to rfc-editor.org instead of datatracker.

larseggert:
This reduces the diff between editors' copies and submitted versions.


martinthomson:
Yes, the datatracker links are not good.

My only reservation here is that the rfc-editor.org site is - if anything - slower than dt.ietf.org.

PR #318, opened 2021-11-25
The markdown table has too many column headers.

mnot:


build-index.sh
```diff
@@ -117,8 +117,8 @@ elif [[ "$format" = "md" ]]; then
         echo
     }
     function table_i() {
-        echo "| Draft |     |     |     |     |     |     |"
-        echo "| ----- | --- | --- | --- | --- | --- | --- |"
+        echo "| Draft |     |     |     |     |"
+        echo "| ----- | --- | --- | --- | --- |"
```
martinthomson, line 121: This just highlights a problem...

The main branch has this many, but other branches have one more.

I've been considering dropping the issue link for branches other than the main one.  That might make this easier.  WDYT?

mnot: I think dropping `issues` as well as `datatracker` makes lot of sense. One might even argue that `diff with last submission` isn't relevant to branches...

martinthomson:
Sounds like a plan.  The top entry will have all the good stuff, the other won't.

PR #325, opened 2022-04-18
Mark the GITHUB_WORKSPACE as safe

divergentdave:
Today's automated build of this action's containers pulled in the latest version of Git, which includes a fix for CVE-2022-24765. See https://github.blog/2022-04-12-git-security-vulnerability-announced/. As a result, the Github Actions action is failing when it executes git commands inside the container, with the following error message.

```
fatal: unsafe repository ('/github/workspace' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /github/workspace
```

This PR fixes the issue by running such a command inside the container's entry point script, as recommended here https://github.com/actions/checkout/issues/766.

### DIFF
--- a/docker/action/entrypoint.sh
+++ b/docker/action/entrypoint.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -e
+git config --global --add safe.directory "$GITHUB_WORKSPACE"
 if [ "$1" = "setup" ]; then
   echo "Running setup"
   exec make -f lib/setup.mk


divergentdave:
I did a test-run using my fork, and the action succeeded. Here are the logs: https://github.com/divergentdave/ppm-specification/runs/6064820766

dconnolly:
Yeah this is biting us too: https://github.com/cfrg/draft-irtf-cfrg-frost/runs/6066077574?check_suite_focus=true

SpencerDawkins:
I'm not asking anyone to hurry in **applying** changes to widely used IETF software, but given that there's a PR for this problem, is there an **obvious ETA for the PR to be reviewed and, potentially, merged**?

larseggert:
It's @martinthomson's repo, I assume he will get to it soon. (We did just have a long holiday weekend in many parts of the world.)

SpencerDawkins:
@larseggert - thanks! I was thinking that it was nighttime for @martinthomson, but was forgetting that Monday was also a holiday in many places.

Like I said - I didn't intend to prod anyone, I was just checking for information.

martinthomson:
Sorry that I missed this folks. Vacation.

Why the entrypoint script and not the Docker image? I think that the base image is probably a better choice for this sort of command.

divergentdave:
The value of `GITHUB_WORKSPACE` is provided at runtime, and while my understanding is that it consistently points to a mount at `/github/workspace`, I think hard coding that path would be fragile.

PR #332, opened 2022-05-31
Fix GitHub pages index when using mmark.

dwaite:
Building of the index.html file for GitHub Pages is broken with the
`mmark` tool due to the metadata being in TOML rather than YAML format.

This results in an index being generated without any extracted metadata.

This change adds the TOML tool and differentiates by the header
characters ('%%%' vs '---') which format to parse.

### DIFF
--- a/doc/SETUP.md
+++ b/doc/SETUP.md
@@ -127,10 +127,10 @@ these, I set `GOPATH=~/gocode`.
 
 ## Other tools
 
-If you wish to build GitHub Pages, you will also need to install the `toml` Python package:
+If you wish to build GitHub Pages, you will also need to install the Python packages `yaml` (for kramdown-rfc Markdown) and/or `toml` (for mmark Markdown):
 
 ```sh
-$ pip3 install --user toml
+$ pip3 install --user toml yaml
 ```
 
 Some other helpful tools are listed in `config.mk`.
--- a/doc/SETUP.md
+++ b/doc/SETUP.md
@@ -96,7 +96,6 @@ on the first line of the file:
 
 * `mmark` files must start with '%%%'
 
-
 ### kramdown-rfc
 
 [`kramdown-rfc`](https://github.com/cabo/kramdown-rfc) requires
@@ -128,8 +127,13 @@ these, I set `GOPATH=~/gocode`.
 
 ## Other tools
 
-Some other helpful tools are listed in `config.mk`.
+If you wish to build GitHub Pages, you will also need to install the `toml` Python package:
 
+```sh
+$ pip3 install --user toml
+```
+
+Some other helpful tools are listed in `config.mk`.
 
 # Update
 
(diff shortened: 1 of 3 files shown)


doc/SETUP.md
```diff
@@ -128,8 +127,13 @@ these, I set `GOPATH=~/gocode`.
 
 ## Other tools
 
-Some other helpful tools are listed in `config.mk`.
+If you wish to build GitHub Pages, you will also need to install the `toml` Python package:
```
martinthomson, line 130:
```suggestion
If you wish to build GitHub Pages from markdown, you will also need to install
the `yaml` (for kramdown-rfc) and `toml` (for mmark) Python package:
```

I think that this has to be an "and" to avoid having the script complain when you invoke it.  You could conditionally import `toml` or `yaml` in the script, but this seems easiest.

doc/SETUP.md
```diff
@@ -128,8 +127,13 @@ these, I set `GOPATH=~/gocode`.
...
+
+```sh
+$ pip3 install --user toml
```
martinthomson, line 133:
```suggestion
$ pip3 install --user toml yaml
```

extract-metadata.py
```diff
@@ -7,16 +7,29 @@
...
 import yaml
-
+import toml
 
 def extract_md(filename):
     try:
         with open(filename, "r") as fh:
-            return next(yaml.safe_load_all(fh))
+            section_header = fh.readline().strip()
+            if (section_header == r'%%%'):
```
martinthomson, line 16:
```suggestion
            if section_header == r"%%%":
```

Would you mind running [black](https://pypi.org/project/black/) over this script?

extract-metadata.py
```diff
@@ -7,16 +7,29 @@
...
+                header_data = ""
+                for line in fh:
+                    if line.strip() == r"%%%":
+                        break
+                    header_data += line
+                return toml.loads(header_data)
+            elif (section_header == r'---'):
+                return next(yaml.safe_load_all(fh))
+            else:
+                raise Exception("Unexpected file section header at top of file \"{0}\": expected `\%\%\%` or `---` ".format(section_header))
```
martinthomson, line 26:
```suggestion
                raise Exception(f"Unexpected first line in markdown file \"{section_header}\": expected `\%\%\%` or `---`")
```
f-strings ftw

martinthomson (commented):
Much appreciated.  A few tweaks suggested, but it would be good to fix this.

doc/SETUP.md
martinthomson: Maybe just this instead:

```suggestion
If you wish to build GitHub Pages, you will also need to install the `toml` and `yaml` Python packages:
```

dwaite:
I tried to conditionally import and add more prescriptive errors, however this will be eaten by the current function. I can submit a change to output exceptions to stderr and emit a failure return code in main as a separate PR.

PR #333, opened 2022-05-31
Generated index to use title (instead of abbrev)

dwaite:
Use the more descriptive and appropriate full title for the specification in the generated index, rather than the abbreviation.

### DIFF
--- a/build-index.sh
+++ b/build-index.sh
@@ -194,7 +194,7 @@ function list_dir() {
 
         tr_i
         src=$(ls "$file".{md,xml} 2>/dev/null | head -1)
-        title=$("${libdir}/extract-metadata.py" "$src" abbrev)
+        title=$("${libdir}/extract-metadata.py" "$src" title)
         td "$(a "$(reldot "$dir")/${file}.html" "${title}" html "$file")"
         td "$(a "$(reldot "$dir")/${file}.txt" "plain text" txt "$file")"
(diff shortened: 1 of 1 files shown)


dwaite:
The behavior was changed in September, 2021 . From the terse commit message, I can't tell if complete omission of the title was intended change.

https://github.com/martinthomson/i-d-template/commit/9d741431a2f9f2037cd0d8dee8ea1c2dc2117970

Perhaps the proper fix, if there are projects for which either a terse or more fully specified list is needed, is to move styling to a stylesheet which is copied in. Then, a repo stylesheet (if present) could override the default shipped stylesheet.

martinthomson:
The change *was* deliberate.  It was to support displays like https://httpwg.org/http-extensions/ where a longer name in the table is onerous.  Are there examples where the abbreviated title doesn't work or ends up being misleading?  I realize that we should perhaps put a `title` attribute with the full title on the link as well.  Would that be helpful?

dwaite:
As generation of an index is more of a bike-shedding issue, I'm thinking of adding functionality to generate a JSON metadata file. That would allow another action in the pipeline to overwrite/generate whatever HTML index they desired.

martinthomson:
If you work out what you want, I'm happy to help facilitate.  Creating extra files on gh-pages isn't that much of a big deal.

Issue #334, opened 2022-06-09
Don't rely on user-installed tools


mnot:
For command-line usage, it'd be more reliable to set up a .venv for the Python tools, and for Ruby I think the equivalent is RVM.

For Python, I highly recommend:
  https://github.com/sio/Makefile.venv

martinthomson:
Most of these changes are on a branch (mnot-334).  I'm fairly confident in them now (I've been using them myself).  However, given the size of the changes and the upcoming IETF meeting, I'm going to wait until the submission deadline passes to land them.

mnot:
This can be closed, no?

martinthomson:
Yes, I think so.  Thanks for the reminder.

PR #335, opened 2022-06-10
Mnot/334

mnot:
Addressing #334.

TODO:

- [x] cache .venv in CI
- [x] test archiving
- [x] `toml` and `pyyaml` for GitHub pages (along with appropriate changes to those jobs)
- [ ] Find other places that use Python and redirect them to venv
- [x] Ruby


archive.mk
```diff
@@ -39,7 +39,7 @@ archive.json: fetch-archive $(drafts_source)
 	old_archive=$$(mktemp /tmp/archive-old.XXXXXX); \
 	trap 'rm -f $$old_archive' EXIT; \
 	git show $(ARCHIVE_BRANCH):$@ > $$old_archive || true; \
-	$(trace) archive -s archive-repo $(python) -m archive-repo archive $(GITHUB_REPO_FULL) $(GITHUB_API_TOKEN) $@ --reference $$old_archive;
+	PYTHONPATH=$(VENV) $(trace) archive -s archive-repo $(VENV)/python -m archive-repo archive $(GITHUB_REPO_FULL) $(GITHUB_API_TOKEN) $@ --reference $$old_archive;
```
martinthomson, line 42: The best way to manage this is to `export PYTHONPATH`.

Also, I would define `python := $(VENV)/python`

requirements.txt
```diff
@@ -0,0 +1,2 @@
+xml2rfc==3.12.10
```
martinthomson, line 1: Hmm, does this really need to be tied to a specific version?  I know it is good practice, but it's a pain in the butt.

requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+archive-repo==0.0.2
```
martinthomson, line 2: Add:

svgcheck
toml
pyyaml

Also, I wonder if we should also look in ./requirements.txt as well (if it exists) so that people who need extra tools can just make their own.  Or is that in Makefile.venv in a way that is awkward?

martinthomson (commented):
Don't forget .gitignore.

archive.mk
mnot: defining `python` like that doesn't work in some places, e.g., `build-index.sh`.

requirements.txt
mnot: what do you mean by 'we should also look'? makefile.venv installs them automatically.

requirements.txt
mnot: Also, if we don't include the version, it's going to make CI cache file updating problematic

requirements.txt
martinthomson: kramdown runs it if you include an SVG using any of the 20-ish ways that it supports.  We don't need to deal with all of those ways here (those can stay in the system), but it is a common-enough (and small enough) addition I thought it worthwhile.

mnot: Hmm, OK. I assume it needs to be on the path when kramdown-rfc2629 is run?

mnot:
Arg, I just realised that `requirements.txt` and `Gemfile` will need to both be in the template so that they're available when the CI cache is inflated. @martinthomson is that odious? I can perhaps make them dot files, but they'll still be at the top-level index of the repo...

requirements.txt
mnot: Perhaps what we can do is list `xml2rfc>=3.12.10` (and similar for other deps), so the latest will be installed, but we can still bump it up once in a while to assure that old installations get upgraded.

martinthomson:
> requirements.txt and Gemfile will need to both be in the template 

I think that we can be dumber.  As I mentioned, we can put a venv directory in the docker image and then copy it into place.  It means that we don't get updates dynamically, we have to wait on a new image, but it means that we don't rely on the cache and all that.  Because, yeah, putting new files into the template is more than a little odious, mostly because I don't think that we can get away with it without a lot more disruption.

mnot:
Can the docker image put `.gems` and `.venv` in the top-level dir of the repo? If so that will work.

mnot:
Actually, we might be overthinking this. If the docker image installs the appropriate libraries in the system, the package cache should have copies available for the venv -- at least for Python. Not sure about Ruby.

mnot:
This needs some testing, but I think it's getting close. Not sure about ruby package caching as per above.

mnot:
still working on an issue in build-index.sh

config.mk
```diff
@@ -49,7 +48,13 @@ XML_RESOURCE_ORG_PREFIX ?= https://xml2rfc.tools.ietf.org/public/rfc
 # This is for people running macs
 SHELL := bash
 
-python ?= /usr/bin/env python3
+# Python venv configuration
+REQUIREMENTS_TXT = $(abspath $(LIBDIR)/requirements.txt)
```
martinthomson, line 52:
```suggestion
REQUIREMENTS_TXT = $(abspath $(LIBDIR)/requirements.txt $(wildcard requirements.txt))
```

config.mk
```diff
@@ -49,7 +48,13 @@ XML_RESOURCE_ORG_PREFIX ?= https://xml2rfc.tools.ietf.org/public/rfc
...
+VENVDIR ?= $(abspath $(LIBDIR)/.venv)
+
+# Ruby bundy configuration
+export BUNDLE_PATH ?= $(abspath $(LIBDIR)/.gems)
+export BUNDLE_PATH_RELATIVE_TO_CWD = true
```
martinthomson, line 57:
```suggestion
```

PR #337, opened 2022-06-23
Fix docker file

aaomidi:
The docker build was failing without these changes


martinthomson:
The build is working: see https://github.com/martinthomson/i-d-template/actions/runs/2519088452

Is this because you are using Julian Reschke's XSLT?

PR #355, opened 2022-12-03
Add plantuml package to GitHub Actions image

JKRhb:
Using the template and its GitHub Actions workflows for rendering an Internet Draft, I noticed that it is currently not possible to include figures containing plantUML code, since the [plantuml package](https://pkgs.alpinelinux.org/package/edge/community/x86_64/plantuml) is not included in the Docker image. This PR adds the package to the Dockerfile, which should enable this feature when writing a draft.

### DIFF
--- a/docker/action/Dockerfile
+++ b/docker/action/Dockerfile
@@ -20,6 +20,7 @@ RUN set -e; \
       nodejs \
       npm \
       openssh \
+      plantuml \
       py3-appdirs \
       py3-configargparse \
       py3-html5lib \


martinthomson:
Hi @JKRhb, thanks for the pull request.  I'm seeing two issues with this that might make this harder to accept than otherwise.

1. This doesn't build for me.  I see the package [here](https://pkgs.alpinelinux.org/package/edge/community/x86_64/plantuml), but I was not able to build it.

```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.15/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  plantuml (no such package):
    required by: world[plantuml]
```

2. Image size.  An important consideration for performance is image size.  Though this package is small, it also depends on Java packages, which are pretty large.  I don't know exactly how large this ends up being (see problem 1), but based on a rough check it looks like it will basically double the image size.

Have you tried mermaid?  That can be installed simply by adding a `package.json` file to your own repository.

JKRhb:
Thank you for your detailed response, @martinthomson! These are good reasons for not including plantuml in the image, especially considering that it can't be installed/selected for the image :/ Sorry for not trying this out myself before opening the PR.

I just tried using mermaid, there seems to be a problem, though, when trying to use it in a GitHub Actions workflow (see [here](https://github.com/namib-project/draft-coap-mud/actions/runs/3615892091/jobs/6093348852)). The error message [hints at a Node 14 problem with puppeteer](https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-is-downloaded-but-fails-to-launch-on-nodejs-14) that leads to an error when trying to start a Chromium process. However, the currently installed version is 18, right?

All in all, I guess we can just close this PR, and I'll might experiment with mermaid a bit more :)

JKRhb:
With https://github.com/martinthomson/i-d-template/pull/417 merged, I suppose this PR can now definitely be closed as superseded :)

PR #357, opened 2022-12-29
Add CSS rules to fix plantuml SVG in dark mode

snorp:
Add CSS rule to fix plantuml SVG in dark mode

### DIFF
--- a/v3.css
+++ b/v3.css
@@ -995,3 +995,8 @@ svg :is([fill="black"], [fill^="#000"], :not([fill])) {
 svg :is([stroke="fill"], [fill^="#fff"]) {
   fill: var(--background-color);
 }
+
+svg :is([fill="white"], [stroke="black"]) {
+  fill: var(--background-color);
+  stroke: var(--text-color);
+}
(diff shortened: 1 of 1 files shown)


v3.css
```diff
@@ -995,3 +995,8 @@ svg :is([fill="black"], [fill^="#000"], :not([fill])) {
 svg :is([stroke="fill"], [fill^="#fff"]) {
   fill: var(--background-color);
 }
+
+svg :is([fill="white"], [stroke="black"]) {
+  fill: var(--background-color);
+  stroke: var(--text-color);
```
martinthomson, line 1001: I think that this is providing a fill rule for white, which is good, but the stroke rule here is off.  I think that what you want is:

```suggestion
svg :is([fill="white"], [fill="#fff"]) {
  fill: var(--background-color);
```

The stroke rule is already there on [line 986](https://github.com/martinthomson/i-d-template/pull/357/files#diff-106ec305e4dacd3b1b5525f2c9c6988499c19e1b013c89946dfbe209b903cc43R986).

martinthomson (changes requested):

martinthomson:
Thanks!

PR #360, opened 2023-01-09
Use the `deps` target instead of `$(DEPS_FILES)`

snorp:
This allows one to add additional deps in the project Makefile using a
double-colon rule, similar to how we do it for `lint`.

### DIFF
--- a/main.mk
+++ b/main.mk
@@ -68,7 +68,7 @@ ifeq (true,$(TIDY))
 MD_POST += | $(trace) $@ -s tidy $(rfc-tidy)
 endif
 
-%.xml: %.md $(DEPS_FILES)
+%.xml: %.md deps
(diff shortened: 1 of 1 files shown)


martinthomson (changes requested):
This introduces an interesting and subtle problem.  When you do this try running `make` twice and see what happens.  You will find that it rebuilds everything, even when the source files haven't changed.

The problem is that a dependency on a `.PHONY` target is [always run](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html):

> A phony target should not be a prerequisite of a real target file; if it is, its recipe will be run every time make considers that file.

Thus, the right way to add dependencies is to both add to the `deps::` dependency AND to extend `DEPS_FILES` with `+=`.

PR #362, opened 2023-01-27
Switch from asciitex to utftex

larseggert:
utftex is a newer effort that is maintained and produces better (=UTF8) output.

You probably should wait until @cabo has merged
https://github.com/cabo/kramdown-rfc/pull/186


larseggert:
Hang on, need to rework this a bit.

martinthomson:
Given the changes in kramdown-rfc, why not just install both?  It's not like the build time or image size is something we need to protect here.

larseggert:
That's what I'm planning on doing, and also want to install the 1.15 release - but that is b0rked at the moment.

larseggert:
@martinthomson I think this is ready to go. CC @cabo.

PR #365, opened 2023-01-31
Use head from utftex instead of specific release

larseggert:


martinthomson (approved):
Living on the edge, I like it.

PR #392, opened 2023-06-16
updating the I-D link and using "I-D"

ghwood:
Suggestions to use "Internet-Draft" and "I-D" as the standard name for the documents created.


martinthomson:
Thanks Greg.

Issue #416, opened 2023-12-01
Is mermaid-cli supported or would it be?


aj-stein-nist:
Hello, I am working on drafts with the IETF SCITT WG and we wanted to evaluate the use of Mermaid for at least some of our diagrams, not just aasvg. I noticed you had asked another WG contributor if they had considered Mermaid in https://github.com/martinthomson/i-d-template/pull/355#issuecomment-1336539483.

It would require the installation of necessary tools in the Docker image for this to work in GitHub Actions and you have expressed concerns about keeping the container image small, in this and other issues. I will open a PR but will you consider it inclusion given your comment?

OR13:
In theory it should be possible with:

```
npm i --save @mermaid-js/mermaid-cli;
git add package.json;
git commit -m "Add support for mermaid diagrams"
```

Then adding this markdown to the draft:

```
# mermaid

~~~mermaid
erDiagram
    CUSTOMER ||--o{ ORDER : places
    ORDER ||--|{ LINE-ITEM : contains
    CUSTOMER }|..|{ DELIVERY-ADDRESS : uses
~~~
```

In practice this does not work, because mermaid cli requires pupeteer, which cannot find chrome.

aj-stein-nist:
I am drafting PR right now, it would need to be added to the Docker container upon which the GHA workflows run, and if that is merged and approved we can add the necessary variables to make this work.

> In practice this does not work, because mermaid cli requires pupeteer, which cannot find chrome.

martinthomson:
The way you can choose a different image is to refer to the "@v1m" action.  That refers to a different image.  See `.../docker/math` for the `Dockerfile` for that.

aj-stein-nist:
OK so I would still need to install Chromium in that, right? I rebased that PR apologies for the delay in #417. It is now just the commit below targeting the `docker/math/Dockerfile` deps build for re-review.

https://github.com/aj-stein-nist/i-d-template/commit/7fc3237d16dbfac2ca2d3cf3cc2b665e60230221

OR13:
yeah, this should def be in the larger image.

I would assume that if the larger image had chromium this would be all that is needed:

https://github.com/martinthomson/i-d-template/issues/416#issuecomment-1836866786

martinthomson:
Give it a go.  The large image was updated when @aj-stein-nist's pull request was merged.

aj-stein-nist:
[...]

PR #417, opened 2023-12-01
Update Dockerfile to install Chromium for Puppeteer support

aj-stein-nist:
Per #416, adding these deps to the container image to install Chromium and its transient dependencies would facilitate using Puppeteer in GitHub Actions, per the documentation at the address below.

https://pptr.dev/troubleshooting/

The i-d-template maintainer and others recommended mermaid-cli as a viable alternative to aasvg and other tools in #355. The former uses headless Chromium to operate.

### DIFF
--- a/docker/math/Dockerfile
+++ b/docker/math/Dockerfile
@@ -27,4 +27,11 @@ RUN set -e; \
     rm -rf "$tmp"; \
     apk del .locale_build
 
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

martinthomson:
Time to test in production!

PR #440, opened 2024-07-19
Make tools from tool_install executable too

### LINKED ISSUE #439: `tool_install` installs tool executable but does not mark it executable
Hello, I frequently use the Docker container locally, not just in GitHub Actions, for local development and review of documents. What a great tool, thanks for your efforts. I noticed that, even though `idnits` is installed correctly through the Dockerfile and the inline `tool_install` function, unlike the `install` function it does not make the resulting tool executable. Of course, manually using `chmod +x` in the container fixes that, but I would like to push a fix upstream.

```sh
$ docker run -it -v $(pwd):/app --entrypoint /bin/bash ghcr.io/martinthomson/i-d-template-math:latest
47701e80de0f:/app# cd /app
47701e80de0f:/app# make
47701e80de0f:/app# make check
echo versioned/draft-ietf-cbor-cddl-more-control-00.txt | xargs -n 1 sh -c 'lib/trace.sh "$0" -s idnits idnits "$0"'
lib/trace.sh: line 14: /usr/local/bin/idnits: Permission denied
lib/trace.sh: line 14: exec: /usr/local/bin/idnits: cannot execute: Permission denied
make: *** [lib/main.mk:148: idnits] Error 123
```

PR incoming, I will reference this issue.

aj-stein-nist:
Fixes #439, see the issue with bug report and more details.

### DIFF
--- a/docker/action/Dockerfile
+++ b/docker/action/Dockerfile
@@ -64,6 +64,7 @@ RUN set -e; \
       curl -SSLf -o "/usr/local/bin/${tool}" \
         "https://raw.githubusercontent.com/ietf-tools/${tool}-mirror/${version}/${tool}"; \
       [ $(sha256sum -b "/usr/local/bin/${tool}" | cut -d ' ' -f 1 -) = "$sha" ]; \
+      chmod 755 "/usr/local/bin/${tool}"; \
     }; \
     set -x; \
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Ah, it seems like the source files are no longer being correctly produced.  I'll also open a bug there (so that your fix is redundant).

PR #443, opened 2024-08-26
Fix variable reference

MikeBishop:
The variable `$email` doesn't appear to be defined, and this breaks. Changing it to `$tagger` appears to work, but I want to make sure I'm not breaking a different path.

### DIFF
--- a/get-email.sh
+++ b/get-email.sh
@@ -30,7 +30,7 @@ draft="$2"
 for k in taggeremail committeremail authoremail; do
     tagger="$(git tag --list --format '%('"$k"')' "$tag" | sed -e 's/^<//;s/>$//')"
     ghuser="${tagger%@users.noreply.github.com}"
-    if [ "$ghuser" = "$email" ]; then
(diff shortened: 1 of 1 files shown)


martinthomson:
Thanks Mike, easy error to fix, I hope.

MikeBishop:
Looks like `make upload` finds the right thing now; I assume CI jobs won't pick up the change until the container image gets respun?

martinthomson:
New action runs should pick up changes quickly.  There might be some caching that means that new runs don't pick up things absolutely immediately, but I'd put that in the minutes category, not days.

pmeenan:
Any chance this broke the publication to datatracker flow (or maybe it was relying on the failure before)?

I'm getting a [400 error](https://github.com/httpwg/http-extensions/actions/runs/10584469921) from datatracker from the github actuin in the http-extensions repository that started today and the logs are showing {"error": "No such user: noreply@github.com"}.

PR #452, opened 2024-10-17
Fix draft replaces

DavidSchinazi:
`git tag -l` takes a pattern as parameter, so if we have a bunch of tags like `draft-me-silly-nonsense-03` and `draft-me-silly-nonsense-42`, then `git tag -l draft-me-silly-nonsense` returns nothing but `git tag -l draft-me-silly-nonsense-*` finds them


martinthomson (approved):
Ugh, how did that ever work (I swear that I tested this).  Thanks David.

martinthomson:
I realize now that there is a latent bug here: If you have draft-ietf-foo-00 and draft-ietf-foo-bar-00, that's not going to work out well for you.  That's probably not worth losing sleep over, but maybe there is a simple fix for that too.

DavidSchinazi:
Fixing the latent bug at #453

PR #453, opened 2024-10-17
We are pedants are we not

DavidSchinazi:
https://github.com/martinthomson/i-d-template/pull/452#issuecomment-2418322109

### DIFF
--- a/upload.mk
+++ b/upload.mk
@@ -32,7 +32,7 @@ endif
 	  file="$$(git ls-files "$${1%-[0-9][0-9]}.*")"; \
 	  for last in $$(git log --follow --name-only --format=format: -- "$${file%-[0-9][0-9]}" | \
 		sed -e '/^$$/d' | grep -v draft-todo-yourname-protocol | cut -f 2 | uniq | tail +2); do \
-	    if [ -n "$$(git tag -l "$${last%.*}-*")" | grep -E "$${last%.*}-[0-9]{2}")" ]; then \
+	    if [ -n "$$(git tag -l | grep -E "^$${last%.*}-[0-9]{2}$$")" ]; then \
 	      echo -F; echo "replaces=$${last%.*}"; break; \
 	    fi; \
 	  done; \
--- a/upload.mk
+++ b/upload.mk
@@ -32,7 +32,7 @@ endif
 	  file="$$(git ls-files "$${1%-[0-9][0-9]}.*")"; \
 	  for last in $$(git log --follow --name-only --format=format: -- "$${file%-[0-9][0-9]}" | \
 		sed -e '/^$$/d' | grep -v draft-todo-yourname-protocol | cut -f 2 | uniq | tail +2); do \
-	    if [ -n "$$(git tag -l "$${last%.*}-*")" ]; then \
+	    if [ -n "$$(git tag -l | grep -E "$${last%.*}-[0-9]{2}")" ]; then \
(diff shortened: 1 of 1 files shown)


upload.mk
```diff
@@ -32,7 +32,7 @@ endif
 	  file="$$(git ls-files "$${1%-[0-9][0-9]}.*")"; \
 	  for last in $$(git log --follow --name-only --format=format: -- "$${file%-[0-9][0-9]}" | \
 		sed -e '/^$$/d' | grep -v draft-todo-yourname-protocol | cut -f 2 | uniq | tail +2); do \
-	    if [ -n "$$(git tag -l "$${last%.*}-*")" ]; then \
+	    if [ -n "$$(git tag -l | grep -E "$${last%.*}-[0-9]{2}")" ]; then \
```
martinthomson, line 35:
```suggestion
	    if [ -n "$$(git tag -l "$${last%.*}-[0-9][0-9]")" ]; then \
```

martinthomson (changes requested):
git is happy to use globs, so this is probably cleaner.

upload.mk
DavidSchinazi, line 35:
```suggestion
	    if [ -n "$$(git tag -l "$${last%.*}-*")" | grep -E "$${last%.*}-[0-9]{2}")" ]; then \
```

upload.mk
```diff
@@ -32,7 +32,7 @@ endif
 	  file="$$(git ls-files "$${1%-[0-9][0-9]}.*")"; \
 	  for last in $$(git log --follow --name-only --format=format: -- "$${file%-[0-9][0-9]}" | \
 		sed -e '/^$$/d' | grep -v draft-todo-yourname-protocol | cut -f 2 | uniq | tail +2); do \
-	    if [ -n "$$(git tag -l "$${last%.*}-*")" ]; then \
+	    if [ -n "$$(git tag -l "$${last%.*}-*")" | grep -E "$${last%.*}-[0-9]{2}")" ]; then \
```
DavidSchinazi, line 35:
```suggestion
	    if [ -n "$$(git tag -l | grep -E "^$${last%.*}-[0-9]{2}$$")" ]; then \
```

upload.mk
```diff
@@ -32,7 +32,7 @@ endif
 	  file="$$(git ls-files "$${1%-[0-9][0-9]}.*")"; \
 	  for last in $$(git log --follow --name-only --format=format: -- "$${file%-[0-9][0-9]}" | \
 		sed -e '/^$$/d' | grep -v draft-todo-yourname-protocol | cut -f 2 | uniq | tail +2); do \
-	    if [ -n "$$(git tag -l "$${last%.*}-*")" ]; then \
+	    if [ -n "$$(git tag -l | grep -E "^$${last%.*}-[0-9]{2}$$")" ]; then \
```
martinthomson, line 35:
```suggestion
	    if [ -n "$$(git tag -l "$${last%.*}-[0-9][0-9]")" ]; then \
```

You might have missed the simpler fix...

DavidSchinazi: I thought about it, but that matches the tag `aaaaaaa-draft-stuff-foobar-03`. Not sure if that matters

martinthomson: I don't think it does.  If `git tag -l a` doesn't match "ab", then this won't match `aaaaa-draft-whatever-09`.

DavidSchinazi: Oh. Duh

PR #454, opened 2024-10-17
Glob match the digits on labels

martinthomson:
Just to be safe after #452.

@dschinazi, I worked it out!


martinthomson:
David has #453.

PR #455, opened 2024-10-22
Fix backupless in-place sed on macOS

DavidSchinazi:
Nothing is ever easy, macOS tooling strikes again! I was investigating why `make update-files` kept breaking for me by dumping a random `CONTRIBUTING.md-e` file into my checkout, and tracked it down to yet another disagreement between GNU sed and BSD sed. On both platforms, operating in-place with a backup file works the same way, so commands like `sed -i~ -e 's/foo/bar/' file.txt` or `sed -i.favicon -e 's/foo/bar/' file.txt` work fine. However, in-place usage without a backup file is done differently:
* on GNU/Linux, you can use `-i` or `-i''` or `--in-place`
* on BSD/Darwin, you can use `-i ''`

And of course, there is no invocation that works on both. For more details, see [StackOverflow](https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux).

I've tested the `setup-readme.sh` portion of this change locally and can confirm that it solves the issue.

### DIFF
--- a/tests/steps/build_commands.py
+++ b/tests/steps/build_commands.py
@@ -75,11 +75,16 @@ def step_impl(context, target, option):
 
 @when("the draft is broken")
 def step_impl(context):
+    import platform
+    if any(platform.system() == p for p in ['Darwin', 'BSD']):
+        sed_no_backup = ["sed", "-i", ""]
+    else:
+        sed_no_backup = ["sed", "-i"]
     with cd(context.working_dir):
         break_this = glob("draft-*.md")[0]
         run_with_capture(
             context,
-            ["sed", "-i", "-e", "s/TODO Security/{{broken-reference}}/", break_this],
+            sed_no_backup + ["-e", "s/TODO Security/{{broken-reference}}/", break_this],
         )
         context.broken_file = break_this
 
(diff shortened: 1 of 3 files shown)


tests/steps/build_commands.py
```diff
@@ -75,11 +75,16 @@ def step_impl(context, target, option):
 
 @when("the draft is broken")
 def step_impl(context):
+    import platform
+    if any(platform.system() == p for p in ['Darwin', 'BSD']):
```
martinthomson, line 79: I think that this works too.

```suggestion
    if platform.system() in ['Darwin', 'BSD']:
```

martinthomson (approved):
Thanks for that.  I guess this skipped the worst because most of this runs in CI.  Of course, not always.  (I used have a sed_i alias somewhere that I use, but now I mostly create backups and just leave them around instead.  `sed -i~` seems to work on macs.)

tests/steps/build_commands.py
DavidSchinazi: Ah right, yeah I was using `any()` because I was initially using `p in platform.system()` instead of `platform.system() == p` but this is simpler

DavidSchinazi:
Yeah I've confirmed that `sed -i~ -e 's/foo/bar/'` works on macOS, so those uses don't need any changes

tests/steps/build_commands.py
DavidSchinazi: Oh and thinking about this more, it returns `FreeBSD` on that platform, so I switched back to any with a substring match

PR #458, opened 2024-11-19
Silence warning on first build

DavidSchinazi:
When a draft is built for the first time, it always prints this error log:
```
Makefile:2: lib/main.mk: No such file or directory
```

Switching `include` to `-include` will silence the error, since it is always expected on first launch.

### DIFF
--- a/template/Makefile
+++ b/template/Makefile
@@ -1,5 +1,5 @@
 LIBDIR := lib
-include $(LIBDIR)/main.mk
+-include $(LIBDIR)/main.mk
 
 $(LIBDIR)/main.mk:
 ifneq (,$(shell grep "path *= *$(LIBDIR)" .gitmodules 2>/dev/null))


template/Makefile
```diff
@@ -1,5 +1,5 @@
 LIBDIR := lib
-include $(LIBDIR)/main.mk
```
martinthomson, line 2: This isn't that simple, sadly.

I do like this change, but I'll need to tweak update.mk to match.  It's a bit involved, sadly.

DavidSchinazi: A tragedy in 3 acts
* "oh how hard could it be?"
* _checks update.mk_
* _slowly backs away from keyboard_

martinthomson: 😁  Leave it with me.  I've got a lot of work right now, but I know how to drive that.

DavidSchinazi: Thanks! This is incredibly low impact and priority so no rush :-)

PR #478, opened 2025-04-10
Use a more impossible character in sed commands

bifurcation:
The current `update-venue.sh` script [fails if a working group name has a `','` character in it](https://github.com/hpkewg/hpke-pq/actions/runs/14388598227/job/40350006512).  This is the case, for example, in the proposed [HPKE Publication, Kept Efficient (HPKE) WG](https://datatracker.ietf.org/wg/hpke/documents/).  The script fails because the `sed` patterns in it use `,` as the `s///` delimiter.  This PR replaces the `,` delimiter with `!`, in hopes that that is a more impossible character in WG names.  (At least until the formation of the Panic! at the Disco WG)


martinthomson:
Thanks, I hadn't imagined we'd have that in a name, but then ...

PR #482, opened 2025-05-02
Fix the sed expression to work on macOS.

### LINKED ISSUE #481: The `update-readme` target always updates README.md, even if regenerate is off
This only occurs on macOS because its `sed` implementation is that of BSD.  Using `gsed` from Homebrew corrects this:

```sh
[ "$(sed -ne '1{s/^<!-- regenerate: \([a-z]*\) .*-->$/\1/;t x;s/.*/on/;: x;p;}' README.md)" = off ] 
sed: 2: "1{s/^<!-- regenerate: \ ...": unexpected EOF (pending }'s)
jclarke@JCLARKE-M-MGXV:~/src/git/draft-opsarea-rfc5706bis|main 
⇒  echo $?
1
```

```sh
[ "$(gsed -ne '1{s/^<!-- regenerate: \([a-z]*\) .*-->$/\1/;t x;s/.*/on/;: x;p;}' README.md)" = off ]
jclarke@JCLARKE-M-MGXV:~/src/git/draft-opsarea-rfc5706bis|main 
⇒  echo $?    
0
```

I wonder if simplifying this check using a basic `awk` command wouldn't be a better approach:

```make
update-readme: auto_update | update-gitignore
	$(call regenerate,README.md,[ "$$(awk 'NR==1{print $$3}' README.md)" = off ])
```

This expression works on macOS and Linux (though it's not as robust as the sed).  If you like this, I can PR it.

xorrkaz:
This fixes #481 and enables `update-readme` to skip regenerating the file when `<!-- regenerate: off -->` is set.

### DIFF
--- a/update.mk
+++ b/update.mk
@@ -84,7 +84,7 @@ update-gitignore:
 	fi
 
 update-readme: auto_update | update-gitignore
-	$(call regenerate,README.md,[ "$$(sed -ne '1{s/^<!-- regenerate: \([a-z]*\) .*-->$$/\1/;t x;s/.*/on/;: x;p;}' README.md)" = off ])
+	$(call regenerate,README.md,[ "$$(grep -m 1 '^<!-- regenerate:' README.md | sed -e 's/^<!-- regenerate: \([a-z]*\) .*-->$$/\1/')" = off ])
 
 update-codeowners: | update-gitignore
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Thanks!

PR #485, opened 2025-05-27
Do not explicitly call regenerate from update-files.

### LINKED ISSUE #484: `make update-files` overwrites README
The `make update-files` target, which is suggested if things are out of date, overwrites any README customizations, even if the regenerate flag is off.  This is because the target explicitly calls the regenerate function instead of calling the `update-readme` target.  Should this `update-files` target simply depend on the `update-readme` and `update-codeowners` targets without calling the `regenerate` function directly?  If so, I can PR that.

xorrkaz:
This will then respect the flags not to regenerate README and CODEOWNERS.  Fixes #484.

### DIFF
--- a/update.mk
+++ b/update.mk
@@ -112,8 +112,7 @@ UPDATE_CI := update-ci
 else
 UPDATE_CI :=
 endif
-update-files: auto_update update-gitignore update-makefile $(UPDATE_CI)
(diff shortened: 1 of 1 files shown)


martinthomson (approved):

martinthomson:
Thanks!

PR #507, opened 2026-02-18
Add flag to suppress lint

ekr:
In some cases we have markdown that won't pass lint, and it's too much trouble to fix it.


main.mk
```diff
@@ -1,3 +1,7 @@
+ifneq (true,$(NO_LINT))
+LINT_TARGET=lint
```
martinthomson:
```suggestion
LINT_TARGET := lint
```

martinthomson (approved):

PR #525, opened 2026-06-22
Fix gh-pages preview tables rendering as raw Markdown

mnot:
Fixes two bugs that make stale per-branch preview tables render as literal Markdown (details in #524):

- **`build-index.sh`**: relocate `tr_i` past the `[[ -z "$src" ]] && continue` guard so a draft with no resolvable source emits no row. Previously it left a stray `|` that, for a directory's last entry, swallowed the blank line that terminates the table, causing kramdown to render the block as a paragraph (and SmartyPants to mangle the `---` separators into em/en-dashes).
- **`ghpages.mk`**: add the missing command substitution (`$$(comm …)`) so the obsolete-directory cleanup loop actually runs `comm` instead of iterating its literal arguments. Without it the loop is a no-op and obsolete preview dirs are never pruned.

Verified with kramdown 2.5.2: a two-section sample goes from 1 table + 1 broken paragraph to 2 tables; the cleanup loop goes from removing nothing to correctly listing all obsolete preview dirs.

---
*Diagnosed and authored with Claude Code (Claude Opus 4.8); reviewed and submitted by @mnot.*

### DIFF
--- a/build-index.sh
+++ b/build-index.sh
@@ -245,7 +245,6 @@ function list_dir() {
         dir=$(dirname "$file")
         file=$(basename "$file" .txt)
 
-        tr_i
         src="${branch}:$(git ls-tree --name-only "$branch" -- "$file".md "$file".xml 2>/dev/null | head -1)"
         [[ -n "${src##*:}" ]] || \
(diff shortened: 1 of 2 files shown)


martinthomson (approved):
Thanks Claude.

mnot:
Will pass that on.

